PR #4, opened 2017-10-20
Persist all keyrings with password before creating first key tree)

chikeichan:


chikeichan:
It broke first time flow because it was creating first account without password :)

PR #6, opened 2017-10-25
Update keyring and sig-util lib versions in package.json

danjm:
The versions of `eth-hd-keyring`, `eth-sig-util` and `eth-simple-keyring` are not up to date. In particular, the current versions do not include the changes that introduced the `signTypedData` methods. This PR brings the versions up to date with those specified in metamask's package.json.

@danfinlay @kumavis once merged, can you publish the update to npm? Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eth-keyring-controller",
-  "version": "2.1.2",
+  "version": "2.1.3",
   "description": "A module for managing various keyrings of Ethereum accounts, encrypting them, and using them.",
   "main": "index.js",
   "scripts": {
@@ -36,9 +36,9 @@
     "bip39": "^2.4.0",
     "bluebird": "^3.5.0",
     "browser-passworder": "^2.0.3",
-    "eth-hd-keyring": "^1.2.0",
-    "eth-sig-util": "^1.2.2",
-    "eth-simple-keyring": "^1.1.1",
+    "eth-hd-keyring": "^1.2.1",
+    "eth-sig-util": "^1.4.0",
+    "eth-simple-keyring": "^1.2.0",
     "ethereumjs-util": "^5.1.2",
     "loglevel": "^1.5.0",
     "obs-store": "^2.4.1",


danfinlay:
Version 2.1.2 seems to already have these methods as published. Are you sure you have the latest running locally?

danjm:
I investigated further.

### Summary (tldr)

I believe there is a problem that needs to be solved here, but **it is not urgent** and will resolve itself whenever `eth-keyring-controller`'s dependency versions are updated. **Developers will have to use `npm install`** instead of `yarn` until that happens to ensure that signing typed messages can be done when running locally.

----

### Further info:

Whether a `keyring.signTypedMessage is not a function` error is thrown in metamask depends on how the node_modules are built. Installing with `yarn` will lead to a user getting an error when trying to sign a typed message. `npm install` will set up node_modules in such a way that there is no error.

The consequence of this is that signing typed messages is working in the versions downloaded from the chrome store. I assume this is because these are packaged with npm and not yarn (or perhaps yarn with the appropriate configuration). Meanwhile, a developer that runs locally and installs node modules with yarn will get an error.

The problem is that Version 2.1.2's dependencies - in the versions specified in package.json - do not have `signTypedData`. Meanwhile, the versions of `eth-hd-keyring`, `eth-sig-util` and `eth-simple-keyring` specificied in metamasks dependencies do have `signTypedData`.

`yarn` installs two versions of these dependencies one in the root of `node_modules/` that corresponds with the versions in metamask's package.json file, and one in  `node_modules/eth-keyring-controller/node_modules/` that corresponds with `eth-keyring-controller`'s package.json.

The the versions of `eth-hd-keyring`, `eth-sig-util` and `eth-simple-keyring` found in `node_modules/eth-keyring-controller/node_modules/` after installing with yarn are outdated and do not have `signTypedData`. Therefore, `keyring.signTypedData` in `KeyringController/index.js` is undefined.

`npm install` installs only one version of these dependencies. It is installed in the root of `node_modules/` and corresponds with the versions in metamask's package.json file. In this case,  `keyring.signTypedData` in `KeyringController/index.js` is defined.

danfinlay:
That's weird, I didn't know yarn built that differently than npm. Published as `eth-keyring-controller@2.1.3`.

PR #12, opened 2018-04-05
remove race condition on unlockKeyrings output

silto:
Keyrings weren't available at the return of promise of `unlockKeyrings`, making in difficult to use right after its execution. 
For example `submitPassword(xxx).then(...)` was not possible, waiting for the 'update' event neither, I had to use a timeout to make it work.
Also `submitPassword` has a `this.keyrings = keyrings` (using the return value) which is only saved by the fact that`restoreKeyrings` pushes the unlocked keyrings to a newly created this.keyrings.

hope this helps!

### DIFF
--- a/index.js
+++ b/index.js
@@ -446,8 +446,8 @@ class KeyringController extends EventEmitter {
     .then((vault) => {
       this.password = password
       this.memStore.updateState({ isUnlocked: true })
-      vault.forEach(this.restoreKeyring.bind(this))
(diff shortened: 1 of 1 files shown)


whymarrh:
Thanks for the PR! @danfinlay I've added a test case showing the error as well.

danfinlay:
Published in `eth-keyring-controller@3.0.1 and @2.1.5`.

PR #13, opened 2018-04-11
Remove account nickname logic and state

whymarrh:
Where MetaMask/metamask-extension#3946 represents the desire to remove the account nicknames out of the KeyringController, this PR removes that logic from the class. This is a breaking change, as it removes a bunch of public methods, but it does seem appropriate as the nickname logic is somewhat isolated.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,6 @@ class KeyringController extends EventEmitter {
       isUnlocked: false,
       keyringTypes: this.keyringTypes.map(krt => krt.type),
       keyrings: [],
-      identities: {},
     })
 
     this.encryptor = opts.encryptor || encryptor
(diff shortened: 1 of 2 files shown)


danjm:
Is there another PR to update+add code in metamask-extension so that this does not break current metamask functionality?

whymarrh:
I believe this would still need to be published (with a major version bump) before MetaMask would have it included in the project, but yes, we'll need to update MetaMask separately. We can hold off on merging this PR but I think it's useful to review this and settle on what the KeyringController API should look like going forward (even if just in the short term until a replacement is prepared).

danfinlay:
Published as `eth-keyring-controller@3.0.0`.

PR #16, opened 2018-04-19
Fix #persistAllKeyrings password binding

whymarrh:
This PR fixes the way the password value is bound for calls to `KeyringController#persistAllKeyrings`. The password argument was supposed to take on `this.password` when it was undefined, but `null` was being passed through via `#createFirstKeyTree`.

In the test cases, the call arg history for `#persistAllKeyrings` looked like so:

```
[ [ 'password123', [] ],
  [ 'password123', [ [Object] ] ],
  [ null, [ [Object] ] ],
  [ 'password123', [ [Object] ] ],
  [ 'password123', [ [Object] ] ],
  [ null, [ [Object] ] ],
  [ 'password123', [] ],
  [ 'password123', [ [Object] ] ],
  [ null, [ [Object] ] ],
  [ 'password123', [ [Object] ] ],
  [ 'password123', [ [Object] ] ],
  [ null, [ [Object] ] ] ]
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,7 @@ class KeyringController extends EventEmitter {
   createNewVaultAndKeychain (password) {
     return this.persistAllKeyrings(password)
       .then(this.createFirstKeyTree.bind(this))
+      .then(this.persistAllKeyrings.bind(this, password))
       .then(this.fullUpdate.bind(this))
   }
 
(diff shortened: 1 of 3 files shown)


whymarrh:
Admittedly I broke this with removing the `typeof` check in #13, but this seems like the correct fix.

danfinlay:
Could you open an equivalent PR on the `2.x.x` branch, @whymarrh?

whymarrh:
I think this just a 3.x issue as #13 only went into that version (2.x has `typeof` check instead)

danfinlay:
Published as `3.1.1`.

PR #20, opened 2018-05-15
Only initialize keyring once

### LINKED ISSUE #19: Two keyrings get created in an unintended way
This is another issue that I think does not lead to a bug in the end product, but I think is unintended and could potentially be dangerous when developing the extension further.

On line 160 and onwards:

```javascript
const keyring = new Keyring(opts)
return keyring.deserialize(opts)
.then((accounts) => {
  return this.checkForDuplicate(type, accounts)
})
```

I don't know if you realize that you create a keyring assign it to a variable and then override it immediately (assuming an HD Key).

What I fear is that by accident the wrong seed phrase can be read, like this:

```javascript
const keyring = new Keyring(opts)
// This seed will be overidden by the next line
const seed = keyring.mnemonic
return keyring.deserialize(opts)
.then((accounts) => {
  const seed2 = keyring.mnemonic
  // Not anymore the same as above
  // seed !== seed2
})
```

Why this happens is pretty obvious when reading the `eth-hd-keyring` source code.

It also does the unncecessary work of generating a seed, private key etc. twice.

danfinlay:
Fixes #19

Keyrings were being initialized twice. This could potentially lead to
consumers of this library briefly thinking one mnemonic was being used
before another one was created.

This could potentially be related to https://github.com/MetaMask/metamask-extension/issues/3127

### DIFF
--- a/index.js
+++ b/index.js
@@ -158,10 +158,7 @@ class KeyringController extends EventEmitter {
   addNewKeyring (type, opts) {
     const Keyring = this.getKeyringClassForType(type)
     const keyring = new Keyring(opts)
-    return keyring.deserialize(opts)
-    .then(() => {
-      return keyring.getAccounts()
-    })
+    return keyring.getAccounts()
     .then((accounts) => {
       return this.checkForDuplicate(type, accounts)
     })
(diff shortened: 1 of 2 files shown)


danfinlay:
That's right, `deserialize` is actually used by the constructor to consume the `opts` and instantiate the object. By removing `deserialize`, we only create the object once.

danfinlay:
Published as `eth-keyring-controller@3.1.2`

PR #24, opened 2018-05-29
Fix persistAllKeyrings getting an array passed in as the password

alextsg:
@danfinlay @kumavis @whymarrh


kumavis:
this was the only instance where we had arguments passed in implicity via bind without also binding the password

PR #27, opened 2018-07-11
added removeAccount

brunobar79:
Once https://github.com/MetaMask/eth-simple-keyring/pull/11 gets merged I can update the version of eth-simple-keyring here

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eth-keyring-controller",
-  "version": "3.2.0",
+  "version": "3.3.0",
   "description": "A module for managing various keyrings of Ethereum accounts, encrypting them, and using them.",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "browser-passworder": "^2.0.3",
     "eth-hd-keyring": "^1.2.2",
     "eth-sig-util": "^1.4.0",
-    "eth-simple-keyring": "^1.2.2",
+    "eth-simple-keyring": "github:brunobar79/eth-simple-keyring#remove-account",
```
danfinlay: Ready to be changed to `eth-simple-keyring@1.3.0`.

package.json
```diff
@@ -38,7 +38,7 @@
     "browser-passworder": "^2.0.3",
     "eth-hd-keyring": "^1.2.2",
     "eth-sig-util": "^1.4.0",
-    "eth-simple-keyring": "^1.2.2",
+    "eth-simple-keyring": "^1.3.0",
```
danfinlay: Do we also need to include `eth-trezor-keyring@0.1.0`?

brunobar79: That's probably a good idea. Right now I'm adding the Keyring type in the metamask-controller (https://github.com/MetaMask/metamask-extension/pull/4625/files#diff-fd8e90928ae2cc670d306c0244a894d0R128)

If you're planning to add the multisig keyring here too then I should def. do it.

danfinlay: Oh right, it's included by consumers. That's fine with me.

PR #29, opened 2018-12-10
External account support Metamask issue #5640

r001:
We want to add external account support, which enables MEtamask not to store private keys for an external address, but upon confirm request the user will be presented with a poup modal that displays a QR code with tx data, and enables to enter a signature. Once valid signature entered, the transaction is processed by Metamask the standard way. 

Since keyrings are hidden from metamask-controller, a keyring must use keyring-controller to expose data to ui, since we want to popup a modal when a transaction is initiated from a from-address of an external account. 

This update creates three arrays of objects in KeyringController. 
`extToSign` - is used to store tx data of transactions that need to be externally signed by user. 
`extSigned` - is the storage of tx s that were successfully signed by user
`extCancel` - is the storage of tx s whose external signing was canceled by user.

Added some functions that enable controlled change of the three arrays above. 

This pr is required in order to https://github.com/MetaMask/metamask-extension/pull/6267
work properly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -271,7 +271,7 @@ class KeyringController extends EventEmitter {
   // extCancel contains all the transaction objects that whose signing was canceled by user
   //
   // Returns a Promise
-  upadteExternalSign (payload) {
+  updateExternalSign (payload) {
     try {
         const allowed = ['extToSign', 'extSigned', 'extCancel']
         const filtered = Object.keys(payload)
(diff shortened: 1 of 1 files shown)


kumavis:
Thanks for the work on this! 
not yet certain this is the best approach
lets keep discussion going over in https://github.com/MetaMask/metamask-extension/pull/6267

r001:
This request is outdated.

PR #31, opened 2019-06-09
added optional parameter to addNewAccount()

r001:
These changes are neede for Bidirectional QR feature ([PR #6267](https://github.com/MetaMask/metamask-extension/pull/6267)) to work.
With `addNewAccount()` we want to specify an address when adding new account to the `eth-bidirectional-qr-keyring`, but current implementation does not allow for specifying any parameter.


r001:
@miranda76 We are pretty close to be able to use keyring controller with ["bidirectional qr codes"](https://github.com/MetaMask/metamask-extension/pull/6267). As you have implemented most of the things this pr has requested, there is only a tiny change needed that is included in this pull request. Please review.

whymarrh:
@r001 thanks for putting this up, and sorry about the glacial review.

I want to make sure I understand the approach here.

> The problem with current implementation is that it is not possible to send `txID` or `msgID` to keyring, that would be needed for signing.

> And also `addNewAccount()` we want to specify an address when adding new account to the `eth-bidirectional-qr-keyring`, but current implementation does not allow for specifying any parameter.

These two comments seem to reference different things? Is this description still accurate?

r001:
> @r001 thanks for putting this up, and sorry about the glacial review.
> 
> I want to make sure I understand the approach here.
> 
> > The problem with current implementation is that it is not possible to send `txID` or `msgID` to keyring, that would be needed for signing.

This is already resolved by previous commits, so nothing to be done here anymore.

> 
> > And also `addNewAccount()` we want to specify an address when adding new account to the `eth-bidirectional-qr-keyring`, but current implementation does not allow for specifying any parameter.
> 

This is the only single change I would like to make to the keyring controller. This is critical for adding the feature of [Bidirectional QR PR #6267](https://github.com/MetaMask/metamask-extension/pull/6267) to work. That pr enables to have accounts that neither have locally stored private keys, nor do they have access to hardware wallet. They only store the account address, and if signing is needed they provide QR with tx data, and expect tx signature from external signer. Thus when you add a new Bidirectional QR account with `addNewAccount()`, you need to be able to provide the address of that account, and thus you need the `opts` argument to provide one.

whymarrh:
Thanks for the context, that makes more sense.

Would you mind rebasing this on the latest and, more importantly, adding a few test cases for this change?

Gudahtt:
Closing this for now, since this PR has been stale for years. Sorry for the delay originally in replying to this; feel free to propose it again if you still want this. Hopefully the QR code integration we've since added on the extension fulfills the requirements the referenced extension PR was meant to.

PR #34, opened 2019-08-21
Add getAppKeyAddress() and support for signing with app keys

danjm:
Adds a simple solution for generating app key public addresses and signing with private app keys.

### DIFF
--- a/index.js
+++ b/index.js
@@ -295,7 +295,7 @@ class KeyringController extends EventEmitter {
   // This method signs tx and returns a promise for
   // TX Manager to update the state after signing
 
-  signTransaction (ethTx, _fromAddress, opts) {
+  signTransaction (ethTx, _fromAddress, opts = {}) {
     const fromAddress = normalizeAddress(_fromAddress)
     return this.getKeyringForAccount(fromAddress)
     .then((keyring) => {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -325,20 +325,35 @@ class KeyringController extends EventEmitter {
...
     const address = normalizeAddress(msgParams.from)
     return this.getKeyringForAccount(address)
       .then((keyring) => {
-      return keyring.signTypedData(address, msgParams.data)
+      return keyring.signTypedData(address, msgParams.data, opts)
+    })
+  }
+
+  // returns an app key
+  getAppKeyAddress (_address, origin) {
```
Gudahtt: See here: https://github.com/MetaMask/eth-simple-keyring/pull/16#discussion_r316258931
The suggestions made in that comment apply here as well

index.js
```diff
@@ -296,11 +296,11 @@ class KeyringController extends EventEmitter {
   // This method signs tx and returns a promise for
   // TX Manager to update the state after signing
 
-  signTransaction (ethTx, _fromAddress) {
+  signTransaction (ethTx, _fromAddress, opts) {
```
danfinlay: I'd just like us to keep new params like `opts` optional across the board if possible.

PR #36, opened 2019-08-27
Add typescript declaration file

danfinlay:
Also imported some type dependencies from `eth-sig-util` and
`ethereumjs-tx`, which I added to the `package.json`.

Also added `yarn.lock`.


danfinlay:
@Gudahtt Good catch, change made.

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  getAccounts(): Promise<Array<string>>
+
+  // Key usage
+  signTransaction(tx: Transaction, from:string): Promise<Transaction>
+  signMessage(msgParams: Object): Promise<string>
+  signPersonalMessage(msgParams: Object): Promise<string>
+  signTypedMessage (params: ITypedMessageParams): Promise<string>
+
+  // Internal methods
+  createFirstKeyTree(): Promise<Array<string> | null>
```
Gudahtt: It looks like the return value of this should be `Promise<null>`. It never returns an array of strings, though it does have a side-effect of emitting the first account (if found) as the `newVault` event.

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  addNewAccount(toKeyring: IKeyring): Promise<IKeyringControllerState>
+  exportAccount(address:string): Promise<string>
+  removeAccount(address:string): Promise<IKeyringControllerState>
+  checkForDuplicate(type: string, newAccount: string): Promise<string>
+  getAccounts(): Promise<Array<string>>
+
+  // Key usage
+  signTransaction(tx: Transaction, from:string): Promise<Transaction>
+  signMessage(msgParams: Object): Promise<string>
+  signPersonalMessage(msgParams: Object): Promise<string>
```
Gudahtt: Interesting - the comment for this method states that it returns a buffer, not a string. It looks like the comment is wrong.

The same for `signMessage` as well.

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  submitPassword(password:string): Promise<IKeyringControllerState>
+  addNewKeyring(type:string, opts: Object): Promise<IKeyring>
+  removeEmptyKeyrings(): Promise<void>
+
+  // Account management
+  addNewAccount(toKeyring: IKeyring): Promise<IKeyringControllerState>
+  exportAccount(address:string): Promise<string>
+  removeAccount(address:string): Promise<IKeyringControllerState>
+  checkForDuplicate(type: string, newAccount: string): Promise<string>
+  getAccounts(): Promise<Array<string>>
```
Gudahtt: Confusingly, this method's implementation is in the "Internal methods" section of the file, though it does seem like it ought to be under this section.

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  setLocked(): Promise<IKeyringControllerState>
+  submitPassword(password:string): Promise<IKeyringControllerState>
+  addNewKeyring(type:string, opts: Object): Promise<IKeyring>
+  removeEmptyKeyrings(): Promise<void>
+
+  // Account management
+  addNewAccount(toKeyring: IKeyring): Promise<IKeyringControllerState>
+  exportAccount(address:string): Promise<string>
+  removeAccount(address:string): Promise<IKeyringControllerState>
+  checkForDuplicate(type: string, newAccount: string): Promise<string>
```
Gudahtt: It looks like this method is expecting a string array for `newAccount`, not just a string. It uses `newAccount[0]` as if that's the new account address.

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  createNewVaultAndRestore (password:string, seed: string): Promise<IKeyringControllerState>
+  setLocked(): Promise<IKeyringControllerState>
+  submitPassword(password:string): Promise<IKeyringControllerState>
+  addNewKeyring(type:string, opts: Object): Promise<IKeyring>
+  removeEmptyKeyrings(): Promise<void>
+
+  // Account management
+  addNewAccount(toKeyring: IKeyring): Promise<IKeyringControllerState>
+  exportAccount(address:string): Promise<string>
+  removeAccount(address:string): Promise<IKeyringControllerState>
```
Gudahtt: This signature seems correct, but the comment for the method states that it returns `Promise<void>`

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  // Vault management and security
+  createNewVaultAndKeychain (password:string): Promise<IKeyringControllerState>
+  createNewVaultAndRestore (password:string, seed: string): Promise<IKeyringControllerState>
+  setLocked(): Promise<IKeyringControllerState>
+  submitPassword(password:string): Promise<IKeyringControllerState>
+  addNewKeyring(type:string, opts: Object): Promise<IKeyring>
+  removeEmptyKeyrings(): Promise<void>
+
+  // Account management
+  addNewAccount(toKeyring: IKeyring): Promise<IKeyringControllerState>
```
Gudahtt: The comment for this method seems completely wrong - must be outdated. This is correct.

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type ITypedMessageParams = {
+  from: string // The signing address
+  data: ITypedData
+}
+
+type IKeyringControllerState = {
+  isUnlocked: boolean
+  keyringTypes: Array<string>
+  keyrings: Array<any>
```
Gudahtt: Could this be `Array<IKeyring>` instead?

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  withAppKeyOrigin?: string
+}
+
+export interface IKeyring {
+  serialize(): Promise<any>
+  deserialize(serialized:any): Promise<void>
+
+  addAccounts(numberToAdd: number): Array<string> // returns address array
+  getAccounts(): Promise<Array<string>>
+  removeAccount(address:string): Promise<void>
```
Gudahtt: `eth-hd-keyring` seems to be missing this method. And in `eth-simple-keyring` it returns `void` (e.g. it isn't async)

index.d.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  signTransaction(address:string, tx: Transaction, opts?: WalletOpts): Promise<string>
```
Gudahtt: This should return `Promise<Transaction>`

index.d.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  addAccounts(numberToAdd: number): Array<string> // returns address array
+  getAccounts(): Promise<Array<string>>
+  removeAccount?(address:string): Promise<void>
+
+  // Implementing getPrivateKeyFor allows inheriting eth-simple-keyring for free signing method support.
+  getPrivateKeyFor?(address:string, opts?: WalletOpts): Buffer
+
+  // If private key method above is not provided, these methods will need to be implemented per-implementation.
+  signTransaction?(address:string, tx: Transaction, opts?: WalletOpts): Promise<string>
```
Gudahtt: Shouldn't these still be mandatory? If they're obtained via inheritance, they'd still be on the class.

lazaridiscom:
some comments:

- I am a bit confused with the code-style in this PR with the "I" prefixes looks a bit alienating. As a comparison, the code-style within gaba's https://github.com/MetaMask/gaba/blob/develop/src/keyring/KeyringController.ts is **very clean** (and should possibly be adopted).
- it is preferred to migrate to ts incrementally, ideally file-by-file, applying refactoring techniques whilst keeping a strict "don't touch anything else" rule.
- a migration (even of one file) must not necessarily be 100% in one chunk (can again be incrementally)

danfinlay:
> I am a bit confused with the code-style in this PR with the "I" prefixes looks a bit alienating.

The course I learned typescript from used the `I` prefix as a convention to clarify that this is an interface, not a class, which could be implemented by multiple classes. Maybe we should just drop it.

I agree, could be nicer in chunks. Maybe will break it up when revisiting.

lazaridiscom:
> The course I learned typescript from used the I prefix as a convention to clarify that this is an interface, not a class, which could be implemented by multiple classes. Maybe we should just drop it.

The prefix convention sounds rational, though i've personally an aversion to this "I" prefix. And it looks like the gaba authors have it too.

Possibly the course-designer came from the .NET domain, but I just looked it up, even Microsoft discourages use of "I" prefix:

> Do not use "I" as a prefix for interface names.

https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names

So it looks preferable to drop it

-

As for the incremental thing: I'm talking here out of theory. I'll try to convert a smaller piece of code, to see how it is in practice.

lazaridiscom:
I went through some documents, manuals and experience-reports from projects. It is possible to make the js-to-ts migration gradually/incrementally/refactoring-like, but it needs some research and some trial & errors.

A quite interesting and most possibly challenging task.

lazaridiscom:
@danfinlay, I opened a major task-issue for the overall migration, see: https://github.com/MetaMask/metamask-extension/issues/7601. You can (if you choose so) now safely close this PR.

Is it of priority to convert the KeyringController first? Or asking differently: do you have any packages which you want to "ts" first?

danfinlay:
Because of the order of dependencies, I think the order of MetaMask TypeScript conversion should be:

- `eth-sig-util`
- `keyrings`
- `eth-keyring-controller`
- More of `metamask`

Closing this PR in favor of converting to typescript (not merely adding a declaration file).

PR #37, opened 2019-08-29
Update lockfile

Gudahtt:
There was one dependency change made just before #32 was merged that required a lockfile update


Gudahtt:
This was updated in #38

PR #42, opened 2019-10-10
fix warnings (method deprecated and unused const)

jpgonzalezra:


whymarrh:
Thanks, @jpgonzalezra, we fixed this mostly via #56

PR #45, opened 2019-11-26
Update keyrings and switch to using `getAppKey`

Gudahtt:
Both `eth-simple-keyring` and `eth-hd-keyring` have been updated to `v4.0.0`. This version replaces the `withAppKeyOrigin` option with the `getAppKey` method, which returns the private app key.

Various options have also been removed, if they were used solely to pass along the removed `withAppKeyorigin` option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -295,11 +295,11 @@ class KeyringController extends EventEmitter {
   // This method signs tx and returns a promise for
   // TX Manager to update the state after signing
 
-  signTransaction (ethTx, _fromAddress, opts = {}) {
+  signTransaction (ethTx, _fromAddress) {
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon both https://github.com/MetaMask/eth-simple-keyring/pull/35 and https://github.com/MetaMask/eth-hd-keyring/pull/19 , and v4.0.0 of both packages being published.

I can bump the version of this package as well, once those two are published.

whymarrh:
I'll close this for now while we await MetaMask/eth-simple-keyring#35 and MetaMask/eth-hd-keyring#19

PR #41, opened 2019-10-03
Update index.js

logvik:
we need to provide the interface from eth-simple-keyring

PR #47, opened 2019-11-27
Squashed commit for PR 41

logvik:
We need this feature for https://github.com/MetaMask/metamask-extension/issues/1190.
With this function, we can obtain the encryption public key for the user, that is generated via `nacl` library. Then we can store it and send encrypted messages to the user in a secure manner. The user will be able to decrypt them via "eth_decryptMessage"

https://github.com/MetaMask/KeyringController/pull/41

### DIFF
--- a/index.js
+++ b/index.js
@@ -332,6 +332,34 @@ class KeyringController extends EventEmitter {
     })
   }
 
+  // Get encryption public key
+  // @object address
+  //
+  // returns Promise(@buffer publicKey)
+  //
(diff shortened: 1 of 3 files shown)


lazaridiscom:
Please note this https://github.com/MetaMask/eth-simple-keyring/pull/31#issuecomment-557918902.

danfinlay:
Published in `eth-keyring-controller@5.4.0`

PR #55, opened 2020-03-17
Use yarn

rekmarks:
- uses `yarn`
  - removes `package-lock.json`
  - adds `yarn.lock`
- add `.nvmrc` per `metamask-extension`


rekmarks:
@Gudahtt, ah crap, my bad. I thought it was `yarn migrate`, which doesn't exist, then I convinced myself that just running `yarn` would be sufficient. We can revert if we think it's a problem.

Gudahtt:
Nah I think it's fine 👍 . Re-generating the lockfile is a good thing to do every now and then arguably. Just wanted to make sure it wasn't done accidentally in the future.

PR #56, opened 2020-03-17
Add @metamask/eslint-config

rekmarks:
- **PENDING:** #55 
- Add `@metamask/eslint-config`, related dependencies, and scripts
- Lint
  - Contains a single `eslint-disable` comment in a test, which we can argue about

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,34 @@
+module.exports = {
+
+  root: true,
+
+  parser: 'babel-eslint',
+
+  parserOptions: {
+    'classes': true,
+    'ecmaVersion': 2017,
+  },
+
+  extends: [
+    '@metamask/eslint-config',
+    '@metamask/eslint-config/config/nodejs',
+    '@metamask/eslint-config/config/mocha',
+  ],
+
+  plugins: [
+    'json',
+  ],
+
+  globals: {
+    'window': true,
+  },
+
+  overrides: [{
+    files: [
+      '.eslintrc.js',
+    ],
+    parserOptions: {
+      sourceType: 'script',
+    },
+  }],
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,10 +2,7 @@ module.exports = {
 
   root: true,
 
-  parser: 'babel-eslint',
-
   parserOptions: {
-    'classes': true,
     'ecmaVersion': 2017,
   },
 
--- a/index.js
+++ b/index.js
@@ -613,7 +613,7 @@ class KeyringController extends EventEmitter {
   //
   // Deallocates all currently managed keyrings and accounts.
   // Used before initializing a new vault.
-  clearKeyrings () {
+  async clearKeyrings () { /* eslint-disable require-await: interface compatibility */
     // clear keyrings from memory
     this.keyrings = []
     this.memStore.updateState({
--- a/index.js
+++ b/index.js
@@ -1,15 +1,19 @@
+
+const { EventEmitter } = require('events')
 const log = require('loglevel')
 const ethUtil = require('ethereumjs-util')
-const BN = ethUtil.BN
+
+const { BN } = ethUtil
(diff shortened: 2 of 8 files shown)


index.js
```diff
@@ -593,7 +613,7 @@ class KeyringController extends EventEmitter {
   //
   // Deallocates all currently managed keyrings and accounts.
   // Used before initializing a new vault.
-  async clearKeyrings () {
+  clearKeyrings () {
```
Gudahtt: Hmm. This seems OK in this case, but I think we should disable the rule that prompted this change (`require-await`).

I thought for a moment that this was `eth-simple-keyring`, and it's important for the keyrings to all share the same API, which means they should all be `async` even if they don't actually do anything async (because another keyring might require `async` for that same method).

.eslintrc.js
```diff
@@ -0,0 +1,31 @@
...
+  root: true,
+
+  parserOptions: {
+    'ecmaVersion': 2017,
+  },
+
+  extends: [
+    '@metamask/eslint-config',
+    '@metamask/eslint-config/config/nodejs',
+    '@metamask/eslint-config/config/mocha',
```
Gudahtt: I think this requires `eslint-plugin-mocha`? That package doesn't seem to have been added.

Gudahtt: Oh wait, it was already there 🤦‍♂

test/index.js
```diff
@@ -100,33 +114,46 @@ describe('KeyringController', () => {
...
+        {
+          getAccounts () {
+            return Promise.resolve([4, 5, 6])
+          },
+        },
       ]
 
       const result = await keyringController.getAccounts()
-      assert.deepEqual(result, [1, 2, 3, 4, 5, 6])
+      assert.deepEqual(result, ['0x01', '0x02', '0x03', '0x04', '0x05', '0x06'])
```
Gudahtt: Huh - why did this change? 🤔

test/index.js
```diff
@@ -200,14 +234,18 @@ describe('KeyringController', () => {
...
+  describe('getAppKeyAddress', function () {
+
+    it('returns the expected app key address', async function () {
       const address = '0x01560cd3bac62cc6d7e6380600d9317363400896'
       const privateKey = '0xb8a9c05beeedb25df85f8d641538cbffedf67216048de9c678ee26260eb91952'
-      const keyring = await keyringController.addNewKeyring('Simple Key Pair', [ privateKey ])
+
+      const keyring = await keyringController.addNewKeyring('Simple Key Pair', [privateKey])
       keyring.getAppKeyAddress = sinon.spy()
+      /* eslint-disable-next-line require-atomic-updates */
```
Gudahtt: I'm having a hard time imagining why this line is required. It looks like `require-atomic-updates` only applies to lines with `yield` or `await` in them? Neither keyword is in the next line

test/index.js
rekmarks: Because I changed `assert` to `strict`, and that's what is actually returned 😬

test/index.js
rekmarks: Same, and yet:
```shell
  245:7  error  Possible race condition: `keyringController.getKeyringForAccount` might be reassigned based on an outdated value of `keyringController.getKeyringForAccount`  require-atomic-updates

✖ 1 problem (1 error, 0 warnings)
```

Gudahtt: Hmm. Maybe something to do with the fact that `keyringController` is re-assigned each test in the `beforeEach` hook? 🤔 

I still don't understand this, but maybe this is because the rule doesn't understand how mocha works.

index.js
rekmarks: Disabled the rule inline; I think it's a good rule in general. I considered overriding it in the config to a warning, but that's not really helpful for future generations.

rekmarks:
@Gudahtt @danfinlay sorry, had to re-request because of reverting `clearKeyrings` to `async`.

PR #57, opened 2020-03-18
Fix docstrings

rekmarks:
- Fixes all docstrings in `index.js`
- Felt like I touched hallowed MetaMask history

### DIFF
--- a/index.js
+++ b/index.js
@@ -695,7 +695,7 @@ class KeyringController extends EventEmitter {
 
   /**
    * Update Memstore Keyrings
-   * 
+   *
    * Updates the in-memory keyrings, without persisting.
    */
   async _updateMemStoreKeyrings () {
@@ -705,9 +705,9 @@ class KeyringController extends EventEmitter {
 
   /**
    * Unlock Keyrings
-   * 
+   *
    * Unlocks the keyrings.
-   * 
+   *
    * @emits KeyringController#unlock
    */
   setUnlocked () {
--- a/index.js
+++ b/index.js
@@ -19,12 +19,9 @@ const keyringTypes = [
 
 class KeyringController extends EventEmitter {
 
+  //
   // PUBLIC METHODS
   //
-  // THE FIRST SECTION OF METHODS ARE PUBLIC-FACING,
-  // MEANING THEY ARE USED BY CONSUMERS OF THIS CLASS.
-  //
-  // THEIR SURFACE AREA SHOULD BE CHANGED WITH GREAT CARE.
 
   constructor (opts) {
     super()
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -386,13 +415,26 @@ class KeyringController extends EventEmitter {
...
     return keyring.getAppKeyAddress(address, origin)
   }
 
+  /**
+   * TODO: what exactly does this do?
+   *
+   * @param {string} _address - The Ethereum address for the app key.
+   * @param {string} origin - The origin for the app key.
+   * @returns {string} The app key address. TODO: wut?
+   */
```
rekmarks: Not totally sure what's going on here. Can someone please add a suggestion?

Gudahtt: It looks like this exports a an app key (e.g. the private key portion of it)

See here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1775.md#private-app-key-generation-algorithm-1

rekmarks: Got it, got it. Fixed.

index.js
```diff
@@ -44,47 +42,52 @@ class KeyringController extends EventEmitter {
...
-  //
-  // returns Promise( @object state )
-  //
-  // Destroys any old encrypted storage,
-  // creates a new encrypted store with the given password,
-  // creates a new HD wallet from the given seed with 1 account.
+  /**
+   * CreateNewVaultAndRestore
+   *
+   * @returns {Promise<Object>} A Promise that resolves to the state.
```
Gudahtt: Nit: elsewhere the `@returns` declarations were moved to the bottom of the doc comment I noticed

index.js
```diff
@@ -44,47 +42,52 @@ class KeyringController extends EventEmitter {
...
+   *
+   * Destroys any old encrypted storage,
+   * creates a new encrypted store with the given password,
+   * creates a new HD wallet from the given seed with 1 account.
```
Gudahtt: Should there be a newline between description and parameters? 🤔 We don't seem to be consistent about this

index.js
```diff
@@ -249,9 +263,9 @@ class KeyringController extends EventEmitter {
   }
 
   // Export Account
```
Gudahtt: Missed this one?

index.js
```diff
@@ -320,12 +335,14 @@ class KeyringController extends EventEmitter {
...
-  // returns Promise(@buffer rawSig)
-  //
-  // Attempts to sign the provided @object msgParams.
+  /**
+   * Sign Message
+   *
+   * Attempts to sign the provided message parameters.
+   *
+   * @param {Object} msgParams - The message parameters to sign.
+   * @returns {Promise<Buffer>} The raw signature, if successful.
```
Gudahtt: Nit: is there a reason the `, if successful` was added here, but not on other return descriptions? (e.g. such as the one for `signPersonalMessage`. It is sorta implied when returning a Promise that the thing-in-the-promise won't be returned if the Promise is rejected.

index.js
```diff
@@ -309,7 +324,7 @@ class KeyringController extends EventEmitter {
 
   // SIGNING METHODS
```
Gudahtt: Another missed one?

index.js
rekmarks: I'm not sure, I think it's more readable. Will add.

Gudahtt: I think I prefer it as well

index.js
rekmarks: I was going back and forth re: this. I'm now of the School of Implied Promise Failure.

PR #60, opened 2020-03-20
Fix unlock event emission

rekmarks:
- Emit `unlock` event:
  - Before persisting keyrings (as before)
  - During `submitPassword`, _after_ `this.keyrings` has been set
     - As opposed to during `unlockKeyrings`, before `this.keyrings` has been set


index.js
```diff
@@ -112,6 +115,7 @@ class KeyringController extends EventEmitter {
         }
         return null
       })
+      .then(this.setUnlocked.bind(this))
```
Gudahtt: I noticed this is _before_ `persistAllKeyrings` here, but _after_ it for the initial vault creation.

We should probably stick to either before or after, for consistency?

PR #63, opened 2020-05-18
Add 'lock' event

rekmarks:
- Emit `lock` in `setLocked`
- Add `setLocked` tests
- Don't emit a `true` value with the `unlock` event
  - This value serves no purpose, and is not used in the extension in any event
- Minor cleanup

We require the `lock` event on the keyring controller to eliminate the `publicConfigStore` in the extension. In addition, there's already an `unlock` event on the keyring controller, and there should be a corresponding `lock` event.

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,8 +122,9 @@ class KeyringController extends EventEmitter {
 
   /**
    * Set Locked
+   * This method deallocates all secrets, and effectively locks MetaMask.
    *
-   * This method deallocates all secrets, and effectively locks metamask.
+   * @emits KeyringController#unlock
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async setLocked () {
@@ -133,6 +134,7 @@ class KeyringController extends EventEmitter {
     // remove keyrings
     this.keyrings = []
     await this._updateMemStoreKeyrings()
+    this.emit('lock')
     return this.fullUpdate()
   }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -122,8 +122,9 @@ class KeyringController extends EventEmitter {
 
   /**
    * Set Locked
+   * This method deallocates all secrets, and effectively locks MetaMask.
    *
-   * This method deallocates all secrets, and effectively locks metamask.
+   * @emits KeyringController#unlock
```
whymarrh, line 127:
```suggestion
   * @emits KeyringController#lock
```

test/index.js
```diff
@@ -36,12 +36,48 @@ describe('KeyringController', function () {
...
+      assert.deepEqual(
+        keyringController.keyrings, [],
+        'keyrings should be empty',
+      )
+    })
+
+    it('emits "lock" event', async function () {
+
+      let called = false
+      keyringController.on('lock', () => {
```
whymarrh, line 70: Can we pass a spy as the callback? That would allow us to avoid keeping this state ourselves.

rekmarks: Done for both `lock` and `unlock` event tests in db251f9

index.js
```diff
@@ -721,7 +723,7 @@ class KeyringController extends EventEmitter {
    */
   setUnlocked () {
     this.memStore.updateState({ isUnlocked: true })
-    this.emit('unlock', true)
+    this.emit('unlock')
```
whymarrh, line 726: This is a breaking change, no? It's unfortunate but someone could be using this for something.

PR #66, opened 2020-05-18
v6.0.0

rekmarks:
`npm version minor`


whymarrh:
As noted in https://github.com/MetaMask/KeyringController/pull/63#pullrequestreview-413877574, I think we'll need v6.

rekmarks:
> As noted in #63 (review), I think we'll need v6.

I hate it, but I can't argue with it!

PR #68, opened 2020-07-02
Ensure memStore only updates once in unlockKeyrings

rekmarks:
Previously, `restoreKeyring` called `_updateMemStoreKeyrings` for each restored keyring. This turned out to be harmful in the single instance it was called, in `unlockKeyrings`. In the extension, we had started to expect a single `memStore` update from the keyring controller on unlock, but instead we received one for each keyring in the controller.

This PR ensures that the `memStore` only updates once for each successful call to `unlockKeyrings`. We accomplish this by adding the `_restoreKeyring` helper method, which performs all the work previously done by `restoreKeyring`, except updating the `memStore`.

`unlockKeyrings` now calls `_restoreKeyring`. `restoreKeyring` remains functionally identical.

### DIFF
--- a/index.js
+++ b/index.js
@@ -546,39 +546,46 @@ class KeyringController extends EventEmitter {
     await this.clearKeyrings()
     const vault = await this.encryptor.decrypt(password, encryptedVault)
     this.password = password
-    await Promise.all(vault.map(this.restoreKeyring.bind(this)))
-    this._updateMemStoreKeyrings()
+    await Promise.all(vault.map(this._restoreKeyring.bind(this)))
+    await this._updateMemStoreKeyrings()
     return this.keyrings
   }
 
   /**
    * Restore Keyring
    *
-   * Attempts to initialize a new keyring from the provided
-   * serialized payload.
-   *
-   * On success, returns the resulting keyring instance.
-   *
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -556,7 +557,9 @@ class KeyringController extends EventEmitter {
    * Attempts to initialize a new keyring from the provided
    * serialized payload.
    *
-   * On success, the resulting keyring instance.
+   * On success, returns the resulting keyring instance.
+   *
+   * ATTN: The caller must call _updateMemStoreKeyrings as necessary.
```
Gudahtt, line 562: It seems unreasonable to expect an consumer of this package to call an internal method. Also, this would be a breaking change, and this isn't the most visible way to call attention to this.

I'd suggest instead factoring out the bulk of `restoreKeyring` to an internal `_restoreKeyring` method that omits the `_updateMemStoreKeyrings` call. Then internally `_restoreKeyring` can be used, followed by the `_updateMemStoreKeyrings` call, but the publicly-facing `restoreKeyring` method can remain unchanged.

rekmarks: Yeah, that was sloppy of me. Done in f90bd69

PR #71, opened 2020-07-23
Adds method to verify password

ryanml:
Related to: https://github.com/MetaMask/metamask-extension/issues/8626


ryanml:
Mistake on my part, as discussed in DM will update

PR #76, opened 2020-10-12
Remove unused `getNetwork` constructor option

Gudahtt:
This option was expected to be passed to the constructor, and was set on the `KeyringController` instance, but was never used.


rekmarks (approved):
Nice!

whymarrh (approved):
Good find!

PR #82, opened 2021-04-21
upgrade eth-sig-util to v3.0.1

brad-decker:
Upgrades eth-sig-util to latest version (v3.0.1) to be inline with the dependencies of `eth-simple-keyring` and `eth-hd-keyring`


Gudahtt (approved):
LGTM!

Neither of the breaking changes from v1 to v3 appear to have impacted `normalize`, which is the sole function we use here from `eth-sig-util`.

PR #87, opened 2021-04-26
Add error info for getKeyringForAccount

### LINKED ISSUE #86: Improve error messages
Our mobile team reports that there are a lot of errors that trace back to this line https://github.com/MetaMask/KeyringController/blob/main/index.js#L679 -- we can improve this error message to provide additional details that will be helpful in diagnosing problems. At a minimum including when there is no address provided, when there are no candidates and/or winners, would be beneficial in troubleshooting. cc @andrepimenta

andrepimenta:
The mobile team has seen multiple errors coming from "No keyring found for the requested account." so this PR adds more info to that error in order to help diagnosing the problem.
Sentry error: https://sentry.io/organizations/metamask/issues/2044540268/?project=2299799&referrer=github_integration
Mobile issue: https://github.com/MetaMask/metamask-mobile/issues/2497

Resolves #86

### DIFF
--- a/index.js
+++ b/index.js
@@ -676,7 +676,19 @@ class KeyringController extends EventEmitter {
         if (winners && winners.length > 0) {
           return winners[0][0]
         }
-        throw new Error('No keyring found for the requested account.')
+
+        // Adding more info to the error
+        let errorInfo = 'Error info: '
+        if (!address) {
+          errorInfo += 'Empty address; '
+        }
+        if (!candidates || !candidates.length) {
+          errorInfo += 'Empty candidates; '
+        }
+        if (!winners || !winners.length) {
+          errorInfo += 'Empty winners;'
+        }
+        throw new Error(`No keyring found for the requested account. ${errorInfo}`)
 
       })
   }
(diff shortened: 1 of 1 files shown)


Gudahtt (commented):
The text "No winners" and "No candidates" might end up being more confusing than helpful :thinking: Those terms don't have any meaning to consumers of this library. Maybe it would be more helpful to express these failure cases as three separate errors?

I see three separate scenarios:
* The `address` passed in is invalid/empty
* There are no keyrings
* There are keyrings, but none match the address.

andrepimenta:
> The text "No winners" and "No candidates" might end up being more confusing than helpful 🤔 Those terms don't have any meaning to consumers of this library. Maybe it would be more helpful to express these failure cases as three separate errors?
> 
> I see three separate scenarios:
> 
> * The `address` passed in is invalid/empty
> * There are no keyrings
> * There are keyrings, but none match the address.

I agree 100%! Just fixed, thank you!

Gudahtt (approved):
LGTM!

cppfuns:
Encountered related problems, please take a look at this submission can solve the problem I encountered 
 https://github.com/MetaMask/metamask-mobile/issues/2497#issuecomment-931132409

PR #95, opened 2021-06-25
Remove unused `addGasBuffer` method

Gudahtt:
This method has nothing to do with keyrings, and wasn't used in either of our products. It was included in this repo since the first commit, possibly by accident.

This is a breaking change, so the release that includes this will need to be a major version bump.


rekmarks (approved):
Yeah, wow.

PR #96, opened 2021-06-25
Remove `ethereumjs-util` dependency

Gudahtt:
The dependency `ethereumjs-util` has been removed. It was only for a few trivial things that were easy to replace.


Gudahtt:
This depends upon #95

rekmarks (approved):
Definitively worth it.

PR #97, opened 2021-06-25
Remove unused test files

Gudahtt:
These files were responsible for creating a `ConfigManager`, which was passed into the KeyringController as a constructor parameter. It was ignored - there is no such thing as a config manager anymore.


rekmarks (approved):
Config manager? Never heard of it.

PR #98, opened 2021-06-25
Remove unused `@babel/core` dependency

Gudahtt:
This dependency is not used anywhere. Seemingly it was added for use by `Babelify`, but `Babelify` isn't used here either.


rekmarks (approved):
Great!

PR #99, opened 2021-06-25
Remove `loglevel` dependency

Gudahtt:
This was only used for a single debug log message. It was not important enough to justify including.


rekmarks (approved):
Agreed!

PR #101, opened 2021-06-28
Remove unused `sinon` sandbox

Gudahtt:
The unused sandbox has been removed. Now the default sandbox is being restored after each test rather than the unused sandbox.


rekmarks (approved):
LGTM!

PR #102, opened 2021-06-28
Update `bip39` from v2.4 to v3.0.4

Gudahtt:
The package `bip39` has been updated. The [breaking change made for v3](https://github.com/bitcoinjs/bip39/blob/master/CHANGELOG.md) was the removal of methods we don't use, as well has a change to another method we don't use.


rekmarks (approved):
LGTM!

PR #103, opened 2021-06-28
Remove browser-related test dependencies

Gudahtt:
All test dependencies related to browser support have been removed. This package works just fine in a Node.js environment - we don't need JSDOM or anything like that. The only strict dependency we have upon browser APIs is within the `browser-passworder` package, which uses the `crypto` API (only available in Node.js >v15). This dependency is used as the default encryptor, but we use a custom mock encryptor in unit tests.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,10 +16,6 @@ module.exports = {
     'json',
   ],
 
-  globals: {
-    'window': true,
-  },
-
   overrides: [{
     files: [
       '.eslintrc.js',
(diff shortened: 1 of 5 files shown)


rekmarks (dismissed):
Looks great!

Gudahtt:
@rekmarks Thanks for the review! Just rebased this to resolve conflicts in `package.json` with the other dependency changes that were merged.

rekmarks (approved):

PR #104, opened 2021-06-28
Remove `bluebird`

Gudahtt:
`bluebird` has been removed, so we're now using native Promises everywhere. Using `bluebird` did not seem necessary; native Promises are sufficient for everything we are doing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "node": ">=12.0.0"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon #103

rekmarks (approved):
Goodbye `bluebird`, I barely knew ya.

PR #105, opened 2021-06-28
Update `ethereumjs-wallet` to v1.0.1

Gudahtt:
`ethereumjs-wallet` has been updated to the latest version. The most notable difference is the removal of the last native dependency, `scrypt`. `scrypt` was an optional dependency that would fail upon install every time because it's incompatible with Node.js v12. Now that error will no longer appear.

The exports for `ethereumjs-wallet` have changed because they migrated to TypeScript, so the import has been modified accordingly.


rekmarks (approved):
thank god

PR #106, opened 2021-06-28
Update `sinon` to the latest version

Gudahtt:
`sinon` has been updated from v7 to v11. [The breaking changes include](https://github.com/sinonjs/sinon/blob/master/CHANGELOG.md) dropping support for legacy browsers and Node.js v8, and changes to the timers API (which we don't use).


rekmarks (approved):
LGTM!

PR #107, opened 2021-06-28
Update ESLint config and dependencies

Gudahtt:
All ESLint config and dependencies have been updated to the latest versions available. The `lint:fix` command has been run as well. All lint violations were fixed automatically.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,13 +1,13 @@
 module.exports = {
   root: true,
-  extends: ['@metamask/eslint-config', '@metamask/eslint-config-mocha'],
+  extends: ['@metamask/eslint-config'],
   env: {
     commonjs: true,
   },
   overrides: [
     {
       files: ['test/**/*.js'],
-      extends: ['@metamask/eslint-config-nodejs'],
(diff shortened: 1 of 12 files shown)


.eslintrc.js
```diff
@@ -1,27 +1,13 @@
 module.exports = {
-
   root: true,
-
-  parserOptions: {
-    'ecmaVersion': 2017,
+  extends: ['@metamask/eslint-config', '@metamask/eslint-config-mocha'],
```
rekmarks, line 3: We should only extend the Mocha config in the test file override.

rekmarks (changes requested):
In addition to the inline comment, we should also add our default `.prettierrc.js`.

.eslintrc.js
Gudahtt: Whoops! Good catch. I had actually thought of this and checked the README for this config to see what was recommended, and the example there doesn't use an override. In that draft there was no override yet so I didn't bother. But now that it's here anyway, there's really no reason not to migrate it.

Gudahtt:
Thanks! Both comments have been addressed

rekmarks (approved):
LGTM!

PR #109, opened 2021-06-28
Add `@lavamoat/allow-scripts`

Gudahtt:
`@lavamoat/allow-scripts` has been added to ensure only allowed post-install scripts are run. A `.yarnrc` file has been added to prevent post-install scripts from running automatically.

A basic 'Contributing' section has been added to explain the new `setup` command. It also includes instructions on setup, linting, and testing.

PR #110, opened 2021-06-28
Migrate from CircleCI to GitHub Actions

Gudahtt:
All CI steps have been migrated from CircleCI to GitHub Actions. The CircleCI badge was removed from the README as well. Everything should work the same except that we're no longer collecting HAR archives of the dependency install step.

The GitHub Action workflow was copied from the metamask module template.


Gudahtt:
This depends upon #109

rekmarks (approved):
You love to see it.

PR #124, opened 2021-09-26
add forget Keyring method

soralit:
Fix the issue that currently MetaMask Extension did not update KeyringController state after forget device so that the addresses should be forgot still remained in redux store and displayed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -725,6 +725,7 @@ class KeyringController extends EventEmitter {
    * Deallocates all currently managed keyrings and accounts.
    * Used before initializing a new vault.
    */
+
   /* eslint-disable require-await */
   async clearKeyrings() {
     // clear keyrings from memory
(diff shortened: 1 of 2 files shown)


Gudahtt (commented):
The KeyringController doesn't currently handle any hardware-wallet-specific methods like this one. In the past we've always called these directly from the extension.

I'm guessing that the problem is that `_updateMemStoreKeyrings` doesn't get called after `forgetDevice` is called in the extension. And `removeAccount` can't be used instead either, because the hardware keyrings don't both implement that method for some reason. So there's no public method to call that would "refresh" the current account state.

I don't love the idea of adding keyring-specific methods :thinking: But the only simple alternative I can think of is a "refreshAccounts" method, and I don't love that idea either because it relies upon the caller to understand the internals of this controller and when accounts need to be refreshed.

So I guess we can accept this for now. And in the new keyring API we'll be proposing soon, we can try to solve this problem in a better way.

index.js
```diff
@@ -757,6 +758,20 @@ class KeyringController extends EventEmitter {
...
+   *
+   * Forget hardware and update memorized state.
+   * @param {Keyring} keyring
+   */
+  forgetKeyring(keyring) {
+    if (keyring.forgetDevice) {
+      keyring.forgetDevice();
+      this.persistAllKeyrings.bind(this)();
+      this._updateMemStoreKeyrings.bind(this)();
+    }
```
Gudahtt, line 773: Could you throw an error here, if `forgetDevice` doesn't exist? Otherwise it might appear that it worked.

Gudahtt (changes requested):

index.js
soralit: Cool

soralit: @Gudahtt The code was updated by adding an else.

PR #132, opened 2022-04-19
Bump eth-hd-keyring to latest version

adonesky1:
Bump eth-hd-keyring to latest version (renamed to @metamask/eth-hd-keyring)

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,10 +5,14 @@ const encryptor = require('browser-passworder');
 const { normalize: normalizeAddress } = require('eth-sig-util');
 
 const SimpleKeyring = require('eth-simple-keyring');
-const HdKeyring = require('eth-hd-keyring');
+const HdKeyring = require('@metamask/eth-hd-keyring');
 
 const keyringTypes = [SimpleKeyring, HdKeyring];
 
+const KEYRINGS_TYPE_MAP = {
+  HD_KEYRING: 'HD Key Tree',
+  SIMPLE_KEYRING: 'Simple Key Pair',
+};
 /**
  * Strip the hex prefix from an address, if present
  * @param {string} address - The address that might be hex prefixed.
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -197,6 +201,11 @@ class KeyringController extends EventEmitter {
   addNewKeyring(type, opts) {
     const Keyring = this.getKeyringClassForType(type);
     const keyring = new Keyring(opts);
+    if (type === KEYRINGS_TYPE_MAP.HD_KEYRING) {
```
Gudahtt, line 204: We don't want to do this if `opts` contains a `mnemonic`. This would erase the old SRP and generate a new one :grimacing:

yarn.lock
```diff
@@ -2028,6 +2038,19 @@ ethereumjs-util@^6.0.0:
     rlp "^2.2.3"
     secp256k1 "^3.0.1"
 
+ethereumjs-util@^6.2.1:
```
Gudahtt, line 2041: Nit: There's a compatible range version of this immediately above this. If you delete the above entry and re-run `yarn`, it should combine them.

adonesky1: tried what you recommended here and didn't see any change 🤔

Gudahtt: What did you try? By "delete the above entry" I meant delete lines 2028-2039. The dedupe was successful for me locally.

index.js
Gudahtt: I'm not really concerned about coverage so much as that the change introduced in this PR is tested. This would not be caught by a coverage report - all of these lines were already covered.

yarn.lock
adonesky1: ah I did mis-read that. Thought you meant this entry (starting line 2041). Worked this time.

index.js
adonesky1: Yes sorry I mean **bump** & **improve** coverage!

adonesky1: I can add a test for this particular issue in this PR though!

adonesky1:
@Gudahtt ready for another look when you have a moment!

index.js
adonesky1: done [here](https://github.com/MetaMask/KeyringController/pull/132/commits/8bfcad706a7f0f20905c145277f53916c522cc20)!

Gudahtt (approved):
LGTM!

PR #136, opened 2022-04-25
Fix issue where QR Code Keyring does not expect an empty options object

adonesky1:
Fixes a bug I introduced with KeyringController's interface with QR hardware wallet Keyring 
`@keystonerhq/base-eth-keyring` doesn't appear to be public on github... but if you look in node_modules (`node_modules/@keystonehq/base-eth-keyring/dist/base-eth-keyring.cjs.development.js`) line 300, the `deserialize` method does not expect an empty object as it's `opts` argument.

This issue became apparent in the QR hardware [keyring tests implemented in Controllers](https://github.com/MetaMask/controllers/blob/main/src/keyring/KeyringController.test.ts#L503)

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,10 +198,10 @@ class KeyringController extends EventEmitter {
    * @param {Object} opts - The constructor options for the keyring.
    * @returns {Promise<Keyring>} The new keyring.
    */
-  addNewKeyring(type, opts = {}) {
+  addNewKeyring(type, opts) {
     const Keyring = this.getKeyringClassForType(type);
     const keyring = new Keyring(opts);
-    if (!opts.mnemonic && type === KEYRINGS_TYPE_MAP.HD_KEYRING) {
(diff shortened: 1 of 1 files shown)


Gudahtt:
Which project did you find this dependency in? It appears to be public on GitHub, here: https://github.com/KeystoneHQ/keystone-airgaped-base/tree/master/packages/base-eth-keyring

Gudahtt:
I think I see what you mean: https://github.com/KeystoneHQ/keystone-airgaped-base/blob/c5e2d06892118265ec2ee613b543095276d5b208/packages/base-eth-keyring/src/BaseKeyring.ts#L290

Good catch, this change makes sense to me.

Gudahtt (approved):

adonesky1:
> Which project did you find this dependency in? It appears to be public on GitHub, here: https://github.com/KeystoneHQ/keystone-airgaped-base/tree/master/packages/base-eth-keyring

Lol I don't know why I couldn't find it...

PR #138, opened 2022-05-10
Patch KeyringController to accept seedphrases passed as buffers/arrays of numbers

adonesky1:
Patch KeyringController  such that `createNewVaultAndRestore` accepts `seedphrase` arg as either type string or array of type number, and passes to addNewKeyring as a buffer.

Updates package to use `@metamask` version of `bip39` that also accepts `seedphrase`s passed as buffers to `validateMnemonic`.

`metamask-extension` already implements this patch [here](https://github.com/MetaMask/metamask-extension/blob/develop/patches/eth-keyring-controller%2B6.2.1.patch) & [here](https://github.com/MetaMask/metamask-extension/blob/develop/patches/bip39%2B2.5.0.patch) (to be removed when this is cut into a release and pulled into the extension)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 const { EventEmitter } = require('events');
-const bip39 = require('bip39');
+const { Buffer } = require('buffer');
+const bip39 = require('@metamask/bip39');
 const ObservableStore = require('obs-store');
 const encryptor = require('browser-passworder');
 const { normalize: normalizeAddress } = require('eth-sig-util');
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,5 +1,6 @@
 const { EventEmitter } = require('events');
-const bip39 = require('bip39');
+const { Buffer } = require('buffer');
```
adonesky1, line 2: Honestly a bit confused why this is necessary... but it seems to be for lint? It's just a node method? cc @Gudahtt

GuillaumeRx (changes requested):
Can you add a test to be sure that it supports the array buffer type seedPhrase ?

index.js
Gudahtt: What did the lint error say?

adonesky1:
``` 
103:11  error  'Buffer' is not defined  no-undef
104:11  error  'Buffer' is not defined  no-undef
```

Screen Shot 2022-05-10 at 1 17 31 PM

adonesky1:
@GuillaumeRx added a test here: https://github.com/MetaMask/KeyringController/pull/138/commits/d7375d4464e51f60c915ca04672a81ab1eb8679b

mcmire (approved):
LGTM!

GuillaumeRx (approved):

index.js
Gudahtt: In a Node.js context, Buffer is a global. ESLint is complaining here because we've declared the environment as being `commonjs` in the ESLint config, so it is not aware that the Node.js globals exist.

The globals are set correctly in our Node.js ESLint configuration, but that isn't being used by this library for some reason. So that would be another way to solve this. Not a blocker though, this seems OK for now.

Issue #127, opened 2022-01-24
Improve unit test coverage


Gudahtt:
We should strive for 100% unit test coverage, and to cover all input boundary conditions.

gantunesr:
After the TS refactor the current code coverage is,

```md
----------------------|---------|----------|---------|---------|----------------------------------------------------------------------------------------------
File                  | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                                                            
----------------------|---------|----------|---------|---------|----------------------------------------------------------------------------------------------
All files             |   90.43 |    68.67 |   92.59 |    90.2 |                                                                                              
 KeyringController.ts |    89.3 |    67.09 |   95.83 |   89.18 | 285,305,391-397,421,449,473,493-499,522,542,562,601,606,678,761,819,825,863,891,906,939,1034 
 constants.ts         |     100 |      100 |     100 |     100 |                                                                                              
 index.ts             |     100 |      100 |      50 |     100 |                                                                                              
----------------------|---------|----------|---------|---------|----------------------------------------------------------------------------------------------
```

The code coverage needs to be increased to a 100% and to keep it at that level it's also needed to enable `jest-it-up`.

PR #140, opened 2022-05-17
improve test coverage

adonesky1:
Bump test coverage from:
Screen Shot 2022-05-18 at 5 17 37 PM

To:
Screen Shot 2022-05-18 at 5 18 22 PM

The remaining coverage is proving a bit tricky. Will come back to it!

Partially fixes: https://github.com/MetaMask/KeyringController/issues/127

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -4,7 +4,7 @@ const sigUtil = require('eth-sig-util');
 const normalizeAddress = sigUtil.normalize;
 const sinon = require('sinon');
 const Wallet = require('ethereumjs-wallet').default;
-
+const HdKeyring = require('@metamask/eth-hd-keyring');
 const { KeyringController, keyringBuilderFactory } = require('..');
 const { KeyringMockWithInit } = require('./lib/mock-keyring');
 const mockEncryptor = require('./lib/mock-encryptor');
@@ -19,6 +19,9 @@ const MOCK_ENCRYPTION_DATA = `{"data":"2fOOPRKClNrisB+tmqIcETyZvDuL2iIR1Hr1nO7XZ
 const walletOneSeedWords =
   'puzzle seed penalty soldier say clay field arctic metal hen cage runway';
 const walletOneAddresses = ['0xef35ca8ebb9669a35c31b5f6f249a9941a812ac1'];
+const walletOnePrivateAddress = [
+  'ace918800411c0b96b915f76efbbd4d50e6c997180fee58e01f60d3a412d2f7e',
+];
 
 const walletTwoSeedWords =
   'urge letter protect palace city barely security section midnight wealth south deer';
@@ -143,6 +146,16 @@ describe('KeyringController', function () {
         expect(actualPassword).toBe(password);
       });
     });
+
+    it('should throw error if accounts are not generated correctly', async () => {
+      jest
+        .spyOn(HdKeyring.prototype, 'getAccounts')
+        .mockImplementation(() => Promise.resolve([]));
+
+      await expect(() =>
+        keyringController.createNewVaultAndKeychain(password),
+      ).rejects.toThrow('KeyringController - No account found on keychain.');
+    });
   });
 
   describe('createNewVaultAndRestore', function () {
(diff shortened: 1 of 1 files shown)


mcmire:
Hey sorry I haven't gotten a chance to look at this lately but I will tomorrow!

test/index.js
```diff
@@ -353,6 +381,24 @@ describe('KeyringController', function () {
       // was also removed after removing the account
       expect(keyringController.keyrings).toHaveLength(1);
     });
+
+    it('does not remove the keyring if there are accounts remaining after removing one from the keyring', async function () {
+      // Add a new keyring with one account
+      await keyringController.addNewKeyring('HD Key Tree', {
+        mnemonic: walletTwoSeedWords,
+        numberOfAccounts: 2,
```
Gudahtt, line 389: The comment says that we're adding a keyring with 1 account, but this is asking the keyring controller to generate 2 accounts

test/index.js
```diff
@@ -353,6 +381,24 @@ describe('KeyringController', function () {
       // was also removed after removing the account
       expect(keyringController.keyrings).toHaveLength(1);
     });
+
+    it('does not remove the keyring if there are accounts remaining after removing one from the keyring', async function () {
+      // Add a new keyring with one account
```
Gudahtt, line 386: This comment says that we're adding a keyring with 1 account, but the test is asking the keyring controller to generate 2 accounts.

I'm guessing that the test is correct but the comment is wrong:

```suggestion
      // Add a new keyring with two accounts
```

test/index.js
```diff
@@ -19,6 +19,9 @@ const MOCK_ENCRYPTION_DATA = `{"data":"2fOOPRKClNrisB+tmqIcETyZvDuL2iIR1Hr1nO7XZ
 const walletOneSeedWords =
   'puzzle seed penalty soldier say clay field arctic metal hen cage runway';
 const walletOneAddresses = ['0xef35ca8ebb9669a35c31b5f6f249a9941a812ac1'];
+const walletOnePrivateAddress = [
```
Gudahtt, line 22: Nit: this seems to be a private key, not a private "address". Maybe worth renaming this variable to `walletonePrivateKey`.

test/index.js
```diff
@@ -636,4 +694,80 @@ describe('KeyringController', function () {
...
+    });
+
+    it('signMessage', async () => {
+      const inputParams = {
+        from: walletOneAddresses[0],
+        data: '0x879a053d4800c6354e76c7985a865d2922c82fb5b3f4577b2fe08b998954f2e0',
+        origin: 'https://metamask.github.io',
+      };
+      const result = await keyringController.signMessage(inputParams);
+      expect(result).toBe(
```
Gudahtt, line 726: Nit: for values like this that a reader can't reasonably be expected to be able to read, it might be worth using an inline snapshot. This communicates to the reader that the value was auto-generated, and makes it easier to update if we do ever intend to change it.

Gudahtt (commented):
Thanks, this looks great!

Once fix to a comment, and a couple of suggestions, but otherwise this looks good. Not sure why CI isn't running though.

test/index.js
adonesky1: done [here](https://github.com/MetaMask/KeyringController/pull/140/commits/3b35dceaf4c0a1170f516d19b41d7f5f8bd57c93)

test/index.js
adonesky1: done [here](https://github.com/MetaMask/KeyringController/pull/140/commits/3b35dceaf4c0a1170f516d19b41d7f5f8bd57c93)

adonesky1: whoops: make `account` plural [here](https://github.com/MetaMask/KeyringController/pull/140/commits/f740476e8472d62acda0b28551312e8d04504f0a)

test/index.js
Gudahtt: Nit: maybe this should be plural since it's an array? I had made it singular at first as well in my comment, but then reconsidered and edited the comment to make ti plural.

PR #145, opened 2022-07-20
Provide method for unlocking keyring by decrypted key

darkwing:
This PR introduced the ability to unlock keyrings by key provided by browser-passworder.  This is part of MetaMask's MV3 effort.


darkwing:
Closing as this strategy would not work due to kdf's not being extractable.

PR #146, opened 2022-08-10
Bump Node to v14

bowensanders:
This changes all refs to v12 to point to v14 as v12 is no longer LTS


paulmillr:
maybe also add v18

mcmire (approved):
This LGTM. If we're interested in adding support for v18, maybe we do that in another PR in case there are any breakages we need to address.

bowensanders:
I agree on waiting until the next PR for 18 - i'm experimenting in the other bump to 14 pr's i'm working on with varied success, and I will be out for vacation for a couple weeks and don't want to logjam any projects over these bumps not going out on time if that's ok with everyone!

PR #147, opened 2022-08-11
Provide method for unlocking keyring by decrypted key

darkwing:
Currently just an outline for for unlocking by decrypted key for MV3


index.js
```diff
@@ -170,12 +171,23 @@ class KeyringController extends EventEmitter {
...
+    this.fullUpdate();
+
+    return this.encryptedKey;
+  }
+
+  /* 
+    MV3:  Unlock by encrypted key
+  */
+  async submitEncryptedKey(encryptedKey) {
+    this.keyrings = await this.unlockKeyrings(undefined, encryptedKey);
```
naugtur, line 186: (could) make it two methods in the API to avoid a chance of confusion - passing the correct secret into the wrong slot would result in what seems to be an incorrect secret and waste time debugging

index.js
```diff
@@ -546,14 +559,40 @@ class KeyringController extends EventEmitter {
...
+      // MV3: Since there's a new salt, we need to generate a new encrypted key
+      // for use in the
+      this.encryptedKey = this._generateEncryptedKey(password, salt);
+    } else {
+      const encryptedVault = this.store.getState().vault;
+      if (!encryptedVault) {
+        throw new Error('Cannot unlock without a previous vault.');
+      }
+      // MV3: We can use an existing salt if one exists in the encrypted key
+      [, salt] = encryptedVault.split(VAULT_SEPARATOR);
```
naugtur, line 593: (could) In my (very limited) experience it pays off to hide the logic of extracting bits of string behind a tiny abstraction if you use it in more than one place. If we ever end up in a situation where we need to migrate from one separator to another or something, we're much better off having a test-covered abstraction where we can quickly introduce tests for the separator migration or any complications we have to add to the salt storage.

index.js
```diff
@@ -585,20 +627,73 @@ class KeyringController extends EventEmitter {
...
 
     await this.clearKeyrings();
-    const vault = await this.encryptor.decrypt(password, encryptedVault);
-    this.password = password;
+
+    // MV3: If the separator string is in the vault string, the user has already migrated
+    // from the previous password-only model
+    let vault = null;
+    if (encryptedVault && encryptedVault.includes(VAULT_SEPARATOR)) {
+      const [, salt] = encryptedVault.split(VAULT_SEPARATOR);
```
naugtur, line 642: The logic of checking what sort of vault content we're dealing with and whether it supports a key or a password would also end up in the abstraction

index.js
```diff
@@ -585,20 +627,73 @@ class KeyringController extends EventEmitter {
...
+      await this.persistAllKeyrings(password);
+    }
+
     return this.keyrings;
   }
 
+  // MV3:  Generates the encrypted key
+  _generateEncryptedKey(password, salt) {
+    // TODO: How complicaated do we want this to be?
+    return window.btoa(password + salt);
```
naugtur, line 675: (must) We want it to be a hash of the two, so that plaintext password is not recoverable. 
(could) Starting with the salt might be preferred for some hashing functions IIRC

index.js
```diff
@@ -585,20 +627,73 @@ class KeyringController extends EventEmitter {
...
+  // https://github.com/MetaMask/browser-passworder/blob/d24b66934ab4d39d97ccff9ff0c3d25f86f1a141/src/index.ts#L180
+  _generateSalt(byteCount = 32) {
+    const view = new Uint8Array(byteCount);
+    global.crypto.getRandomValues(view);
+    // Uint8Array is a fixed length array and thus does not have methods like pop, etc
+    // so TypeScript complains about casting it to an array. Array.from() works here for
+    // getting the proper type, but it results in a functional difference. In order to
+    // cast, you have to first cast view to unknown then cast the unknown value to number[]
+    // TypeScript ftw: double opt in to write potentially type-mismatched code.
+    const b64encoded = window.btoa(String.fromCharCode.apply(null, view));
```
naugtur, line 693: It's a random salt, so maybe you can skip all the pain and use `crypto.randomUUID()` instead.
I don't know what the requirements for entrpoy are though. UUID is 122 bits. I'm not qualified to say if that's enough ;)

Gudahtt, line 693: We could use `Buffer.from(...).toString('base64')` here instead of `window.btoa`. That would ensure this library still works in a Node.js context.

Gudahtt: Oh, i see this was taken from `browser-passworder`, which is only meant to be used by browsers.

Any reason this was inlined, rather than calling `this.encryptor.generateSalt`?

index.js
```diff
@@ -546,14 +559,40 @@ class KeyringController extends EventEmitter {
...
+  persistAllKeyrings(password) {
+    if (password && typeof password !== 'string') {
       return Promise.reject(
         new Error('KeyringController - password is not a string'),
       );
     }
 
-    this.password = password;
+    // MV3: Since we also allow persisting without a password, we should require this.encryptedKey
+    if (password === undefined && this.encryptedKey === undefined) {
```
darkwing, line 570: All of our tests are failing on this check, which is problematic because `persistAllKeyrings` is called without a password and without encryption key, from a few different places.

index.js
darkwing: I thought  about this but that method is kinda complicated and so duplicating it would be a maintenance problem.

index.js
darkwing: Done!

index.js
danjm: I believe the problem is that before the changes in this PR, the tests relied on the `this.password = password;` of what used to be line 522 in the `createFirstKeyTree` function

The stack trace for the error in the tests is:
```
  at KeyringController.persistAllKeyrings (index.js:573:9)
      at index.js:239:21
      at Object.<anonymous> (test/index.js:31:5)
```

So it is happening after calling `await keyringController.createNewVaultAndKeychain(password);` on line 31 in the test file, which is in the `beforeEach` function. `createNewVaultAndKeychain` calls `this.createFirstKeyTree(password)`, but passing the password to `createFirstKeyTree` no longer does anything. So...

I think the fix is to pass the password from `createFirstKeyTree` to `this.addNewKeyring` to `this.persistAllKeyrings()`

index.js
```diff
@@ -516,12 +537,39 @@ class KeyringController extends EventEmitter {
...
+
+    let salt = null;
+    if (password) {
+      // MV3: If this is a migration or new password-driven login, we should
+      // create or rotate the salt
+      salt = encryptor.generateSalt();
+
+      // MV3: Since there's a new salt, we need to generate a new encrypted key
+      // for use in the
+      this.encryptedKey = await this._generateEncryptedKey(password, salt);
```
darkwing, line 563: `this. _generateEncryptedKey` returns an `ArrayBuffer`, so we'll need to convert this to string.

index.js
```diff
@@ -548,20 +602,72 @@ class KeyringController extends EventEmitter {
...
+  parseVault(encryptedVault) {
+    const [vault, salt] = encryptedVault.split(VAULT_SEPARATOR);
+    return { vault, salt };
+  }
+
+  // MV3:  Generates the encrypted key
+  async _generateEncryptedKey(password, salt) {
+    const data = new TextEncoder("utf-8").encode(password + salt);
+    const encryptedSha = await crypto.subtle.digest('SHA-256', data);
+    const encryptedKey = new TextDecoder("utf-8").decode(encryptedSha);
```
darkwing, line 662: The result of this function is generating this error:

```
Unexpected non-whitespace character after JSON at position 1425
```

An example encrypted key is `'nhd\x10x1.ۇ\x04bh$K��{ZT3!L\x06�B����"��'`

test/index.js
```diff
@@ -20,6 +22,20 @@ const walletTwoAddresses = [
...
+
+// Shims for globals that don't exist in test environment
+global.crypto = {
+  getRandomValues: () => {
+    return Math.floor(Math.random(1000000));
+  },
+  subtle: {
+    digest: (algorithm, data) =>
+      new Uint8Array(Date.now().toString().split('')).buffer,
+  },
```
darkwing, line 34: At present, a few of the tests are being effected by the fact that I believe `TextEncoder.decode` is working improperly in the tests.  Each time it appears to return an empty string, which it shouldn't be.  So either the shim is off or the result of what I'm shimming `digest` to do is wrong.

test/index.js
```diff
@@ -451,4 +473,79 @@ describe('KeyringController', function () {
...
+
+      // Log the user out
+      await keyringController.setLocked();
+
+      // Attempt to login with an incorrect encryption key
+      const encryptionKey = await keyringController.submitEncryptedKey(
+        'FAKE KEY',
+      );
+
+      expect(encryptionKey).rejects.toThrow();
```
darkwing, line 513: This test is still passing, which it shouldn't.  I think this is affected by the `_generateEncryptedKey` returning what looks like an empty string in the just the tests.

naugtur: There's no way this test could work. 
You're using an async assertion which should be getting a promise as input, but you await that promise above and pass the return value.

Avoid awaiting the  submitEncryptedKey call, assign the promise it returns to a variable and/or put that reference inside expect(_).rejects

test/index.js
```diff
@@ -451,4 +473,79 @@ describe('KeyringController', function () {
...
+
+      const currentKey = keyringController.getEncryptedKey();
+
+      await keyringController.setLocked();
+
+      await keyringController.submitPassword(password);
+
+      const newKey = keyringController.getEncryptedKey();
+
+      expect(currentKey === newKey).toBe(false);
```
naugtur, line 548: if you use a "not equal" assertion here, you'll get both values in test failure output when it fails, instead of "expected true to be false"

test/index.js
```diff
@@ -451,4 +473,79 @@ describe('KeyringController', function () {
...
+
+      const encryptionKey = keyringController.getEncryptedKey();
+
+      // Log the user out
+      await keyringController.setLocked();
+
+      // Attempt to login with an incorrect encryption key
+      const key = await keyringController.submitEncryptedKey(encryptionKey);
+
+      expect(key).toBeTruthy();
```
naugtur, line 527: Is it in any way important (or relevant) that key is truthy?
Seems like this should assert  submitEncryptedKey() doesn't reject the promise

index.js
```diff
@@ -171,6 +181,19 @@ class KeyringController extends EventEmitter {
...
+
+  /* 
+    MV3:  Unlock by encrypted key
+  */
+  async submitEncryptedKey(encryptedKey) {
+    this.keyrings = await this.unlockKeyrings(undefined, encryptedKey);
+    this.setUnlocked();
+    this.fullUpdate();
+
+    return this.encryptedKey;
```
naugtur, line 196: What's the purpose of this being returned? Is it only for tests? 

If so, there's nothing stopping you from making an assertion on `theReferenceTheTestHas.encryptedKey` instead. It's not a private field.

index.js
```diff
@@ -186,7 +209,9 @@ class KeyringController extends EventEmitter {
     if (!encryptedVault) {
       throw new Error('Cannot unlock without a previous vault.');
     }
-    await this.encryptor.decrypt(password, encryptedVault);
+
+    const result = await this.encryptor.decrypt(password, encryptedVault);
```
naugtur, line 213: (nit) the more common spelling seems to be encrypter :shrug:  
https://en.wiktionary.org/wiki/encryptor

index.js
```diff
@@ -186,7 +209,9 @@ class KeyringController extends EventEmitter {
...
+    return result;
```
naugtur, line 214: jsdoc doesn't suggest this would return anything. Why is that needed?

index.js
```diff
@@ -516,12 +540,39 @@ class KeyringController extends EventEmitter {
...
-    this.password = password;
+    // MV3: Since we also allow persisting without a password,
+    // we should require this.encryptedKey
+    if (password === undefined && this.encryptedKey === undefined) {
+      return Promise.reject(
+        new Error(
+          'KeyringController - a password or encryptedKey must exist to persist keyrings',
+        ),
+      );
+    }
```
naugtur, line 556: Throwing inside an async function has the same result of rejecting the promise returned by it. 
Is this an optimization? Was throwing noticably slower?

index.js
```diff
@@ -532,10 +583,16 @@ class KeyringController extends EventEmitter {
       }),
     );
     const encryptedString = await this.encryptor.encrypt(
-      this.password,
+      this.encryptedKey,
       serializedKeyrings,
     );
-    this.store.updateState({ vault: encryptedString });
+
+    const newVault = [encryptedString, VAULT_SEPARATOR, salt].join('');
```
naugtur, line 590: using a template string will sometimes throw

naugtur, line 590: (could) assert both values are not undefined 
If we introduce a bug somewhere in the future it'd be better to throw an error and leave the password encrypted vault as is than to save something like `VAULT:::undefined` accidentally

index.js
```diff
@@ -548,20 +605,72 @@ class KeyringController extends EventEmitter {
...
 
     await this.clearKeyrings();
-    const vault = await this.encryptor.decrypt(password, encryptedVault);
-    this.password = password;
+
+    // MV3: If the separator string is in the vault string, the user has already migrated
+    // from the previous password-only model
+    let vault = null;
+    if (encryptedVault?.includes?.(VAULT_SEPARATOR)) {
+      const { salt, vault: vaultOnly } = this.parseVault(encryptedVault);
```
naugtur, line 620: (nit)
since this codebase is OO style more than it is FP, maybe a simple Vault class with a constructor that can create the vault object by parsing the string or getting both chunks as input would look nicer. Something like the global URL() which can be used to parse URLs or create them from pieces.

It'd expose public fields:
- salt
- vault
- serialized (the string resulting from concatenation or passed in for parsing)
- isSalted (to use instead of `encryptedVault?.includes?.(VAULT_SEPARATOR)` here)

SEPARATOR would not be known outside the Vault class

index.js
```diff
@@ -548,20 +605,72 @@ class KeyringController extends EventEmitter {
...
+
+        log('[unlockKeyrings] *new* encryptedKey is: ', this.encryptedKey);
+
+      } else if (encryptedKey !== undefined) {
+        log('[unlockKeyrings] *old* encryptedKey is: ', this.encryptedKey);
+        this.encryptedKey = encryptedKey;
+      } else {
+        throw new Error(
+          'No way to decrypt a salted vault without a password or encrypted key',
+        );
```
naugtur, line 635: (could) move the error to the top, throw early if both are missing.

index.js
```diff
@@ -548,20 +605,72 @@ class KeyringController extends EventEmitter {
    * @param {string} password - The keyring controller password.
    * @returns {Promise<Array<Keyring>>} The keyrings.
    */
-  async unlockKeyrings(password) {
+  async unlockKeyrings(password, encryptedKey) {
```
naugtur, line 608: :thinking: Why is it called encryptedKey not encryptionKey? It's not like the key itself is encrypted.

index.js
```diff
@@ -548,20 +605,72 @@ class KeyringController extends EventEmitter {
...
+    } else {
+      vault = await this.encryptor.decrypt(password, encryptedVault);
+    }
+
     await Promise.all(vault.map(this._restoreKeyring.bind(this)));
+
     await this._updateMemStoreKeyrings();
+
+    // MV3: If we're provided a password, we should persist keyrings
+    // so that we can either (1) migrate or (2) create a new salt
```
danjm, line 648: I think the below if block should probably go in `submitPassword`, after `unlockKeyrings` resolves

test/index.js
```diff
@@ -20,6 +22,20 @@ const walletTwoAddresses = [
   '0xbbafcf3d00fb625b65bb1497c94bf42c1f4b3e78',
   '0x49dd2653f38f75d40fdbd51e83b9c9724c87f7eb',
 ];
+
+// Shims for globals that don't exist in test environment
+global.crypto = {
```
naugtur, line 27: FYI, available in node 15
https://nodejs.org/api/webcrypto.html#web-crypto-api

test/index.js
```diff
@@ -20,6 +22,20 @@ const walletTwoAddresses = [
...
+  getRandomValues: () => {
+    return Math.floor(Math.random(1000000));
+  },
+  subtle: {
+    digest: (algorithm, data) =>
+      new Uint8Array(Date.now().toString().split('')).buffer,
+  },
+};
```
naugtur, line 35: the shim could use existing Node crypto implementation if you want results that look more like the originals

```suggestion
const nodeCrypto = require('crypto');
global.crypto = {
  getRandomValues: () => {
    return nodeCrypto.randomFillSync(buffer);
  },
  subtle: {
    digest: (algorithm, data) => {
      const hash = nodeCrypto.createHash(algorithm);
      return hash.update(data).digest();
  },
};
```
^ warning: may contain typos

test/index.js
```diff
@@ -451,4 +473,79 @@ describe('KeyringController', function () {
...
     });
   });
+
+  describe('Manifest v3 changes', function () {
+    it('can still verify password', async function () {
+      // Log the user out
+      await keyringController.setLocked();
+
+      // Attempt to verify the password
+      const result = await keyringController.verifyPassword(password);
```
naugtur, line 483: (nit) verifyPassword throws on failure, so the assertion is not reached if the test won't pass.
using async assertions may produce more helpful output from Jest
```suggestion
     await expect(keyringController.verifyPassword(password)).resolves.toBeTruthy()
```

test/index.js
```diff
@@ -451,4 +473,79 @@ describe('KeyringController', function () {
...
+      expect(result).toBeTruthy();
+    });
+
+    it('can still unlock with password after being migrated and locked', async function () {
```
naugtur, line 487: It's unclear to me. Where did the migration happen?
A `before` call would help me understand what's the precondition, even if redundant.

test/index.js
naugtur: OR
this could be stubbed with sinon in a before or beforeEach and random values would no longer be random so they could be used in assertions

test/index.js
```diff
@@ -451,4 +473,79 @@ describe('KeyringController', function () {
...
+      // Log the user out
+      await keyringController.setLocked();
+
+      // Attempt to unlock the keychaing via old password
+      const result = await keyringController.submitPassword(password);
+      expect(result).toBeTruthy();
+
+      // Ensure the new vault value has separator and salt
+      const { vault, salt } = keyringController.store.getState();
+      expect(vault).toContain(':::');
```
naugtur, line 497: 1. testing an implementation detail. 
2. How is ':::' present in the vault field while salt is already available separately?

If you can use a deterministic shim for getRandomValues you could assert that the salt is correct and present where necessary.

test/index.js
naugtur, line 513:
```suggestion
      const controllerPromise = keyringController.submitEncryptedKey(
        'FAKE KEY',
      );

      await expect(controllerPromise).rejects.toThrow();
```

index.js
```diff
@@ -171,6 +181,19 @@ class KeyringController extends EventEmitter {
...
+    return this.encryptedKey;
+  }
+
+  /* 
+    MV3:  Unlock by encrypted key
+  */
+  async submitEncryptedKey(encryptedKey) {
+    this.keyrings = await this.unlockKeyrings(undefined, encryptedKey);
+    this.setUnlocked();
+    this.fullUpdate();
```
naugtur, line 194: I don't know what this part is doing, assuming some signaling that's not relevant to the security or lock/unlock process.

index.js
```diff
@@ -548,20 +605,72 @@ class KeyringController extends EventEmitter {
...
+
+      if (password !== undefined) {
+        this.encryptedKey = await this._generateEncryptedKey(password, salt);
+
+        log('[unlockKeyrings] *new* encryptedKey is: ', this.encryptedKey);
+
+      } else if (encryptedKey !== undefined) {
+        log('[unlockKeyrings] *old* encryptedKey is: ', this.encryptedKey);
+        this.encryptedKey = encryptedKey;
+      } else {
```
naugtur, line 632: This logic is good :+1:

index.js
```diff
@@ -548,20 +605,72 @@ class KeyringController extends EventEmitter {
...
+      const { salt, vault: vaultOnly } = this.parseVault(encryptedVault);
+
+      log('[unlockKeyrings] salt is: ', salt, '; encryptedVault is:', encryptedVault);
+
+      if (password !== undefined) {
+        this.encryptedKey = await this._generateEncryptedKey(password, salt);
+
+        log('[unlockKeyrings] *new* encryptedKey is: ', this.encryptedKey);
+
+      } else if (encryptedKey !== undefined) {
```
naugtur, line 629: This logic is good :+1:

index.js
```diff
@@ -516,12 +540,39 @@ class KeyringController extends EventEmitter {
...
+
+    let salt = null;
+    if (password) {
+      // MV3: If this is a migration or new password-driven login, we should
+      // create or rotate the salt
+      salt = encryptor.generateSalt();
+
+      // MV3: Since there's a new salt, we need to generate a new encrypted key
+      // for use in the
+      this.encryptedKey = await this._generateEncryptedKey(password, salt);
```
naugtur, line 566: We're overriding the encryptedKey field before going through the (**asynchronous**) process of preparing the new vault.
That leaves a possibility (because of bad timing or an error) for the key in this.encryptedKey to not be a key that can open the current vault.

It's not a big deal, worst case scenario it fails and forces the user to supply password again. Might be annoying if happening right after they supplied the password already.

Keeping the key in local variable and setting it in the same synchronous execution as the call to store the new vault would eliminate that gap.


And while we're at it - where/when is the key being saved to survive a restart? Will this flow save it too early? Why is it not saved in the same synchronous execution as the vault it unlocks?

test/index.js
darkwing: Good improvement!

index.js
darkwing: That's fair.

index.js
darkwing: Agreed, but this is an existing practice that doesn't need to be changed in this PR.

index.js
darkwing: This returns:

```
[
      {
        type: 'HD Key Tree',
        data: {
          mnemonic: [Array],
          numberOfAccounts: 1,
          hdPath: "m/44'/60'/0'/0"
        }
      }
    ]
```

I added for debugging purposes, but yes, it can be removed.

index.js
darkwing: Good idea!

test/index.js
darkwing: Getting error `Digest method not supported` :/

index.js
```diff
@@ -548,20 +637,50 @@ class KeyringController extends EventEmitter {
...
   }
 
+  parseVault(encryptedVault) {
+    const [vault, salt] = encryptedVault.split(VAULT_SEPARATOR);
+    return { vault, salt };
+  }
+
+  // MV3:  Generates the encrypted key
+  async _generateEncryptionKey(password, salt) {
+    return toHex(sha256(utf8ToBytes(password + salt)));
```
Gudahtt, line 676: Why are we using `ethereum-cryptography/sha256` here instead of the encryptor?

index.js
```diff
@@ -168,7 +176,24 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async submitPassword(password) {
+    await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
+
+    // MV3: If we're provided a password, we should persist keyrings
```
Gudahtt, line 182: This package is used by mobile and MV2 extension as well, and these changes to the workflow will apply to both. This "MV3:" prefix might be confusing or misleading.

index.js
danjm: @darkwing I know this code came from the branch I pushed up, but you might have missed this comment in our DM with Zb:

> I think the shim could be used for stubbing random salt with predictable values and asserting for it.
Also, getting a dependency instead using a feature in the browser is not only a detriment to maintainability but also performance :shushing_face:

So we should go back to your old implementation that uses the native crypto api directly. But as per Mark's comment, we can probably just use the encryptor itself

index.js
```diff
@@ -168,7 +174,29 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async submitPassword(password) {
+    await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
+
+    // If we're provided a password, we should persist keyrings
```
danjm, line 180: Nit: this comment should be updated because a password is not optional at this point. I might reword this as: `persistAllKeyrings` is here called so that we can either (1) migrate or (2) create a new salt

index.js
danjm: to be more specific, you could use `keyFromPassword(password, salt)`

danjm: But, we might want to consider how we can more fully leverage browser passworder and push more logic down to that

index.js
```diff
@@ -548,20 +641,62 @@ class KeyringController extends EventEmitter {
...
+      throw new Error(
+        'No way to decrypt a salted vault without a password or encrypted key',
+      );
+    }
+
     await this.clearKeyrings();
-    const vault = await this.encryptor.decrypt(password, encryptedVault);
-    this.password = password;
+
+    // If the separator string is in the vault string, the user has already migrated
```
danjm, line 658: I think this comment is supposed to be on line 620?

index.js
darkwing: I tried doing the following:

```
async _generateEncryptionKey(password, salt) {
    //return toHex(sha256(utf8ToBytes(password + salt)));
    const key = await encryptor.keyFromPassword(password, salt);
    const exportedKey = await window.crypto.subtle.exportKey('jwk', key);
    const keyString = new Uint8Array(exportedKey).toString();

    console.log("key is: ", key, exportedKey);
    console.log("keystring is: ", keyString);
    return key;
  }
```

Result being:

AttemptedKeyExport

The string form of extracted key is empty.  I don't think this helps us.

Gudahtt: It's not totally clear what you're trying to do in that snippet, or how it relates to this thread. But that is failing because you're asking for the key to be exported as a JWT, which is an object. Then you're passing the object into the Uint8Array constructor, which isn't meant to accept that sort of object as input. It tries to iterate over it but it's not iterable, giving you an empty arraybuffer, thus the empty string.

PR #148, opened 2022-08-24
Use async/await instead of then

darkwing:
Much easier to read than the mess of `then`'s.

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,7 +104,7 @@ class KeyringController extends EventEmitter {
         : Buffer.from(seedPhrase);
 
     if (typeof password !== 'string') {
-      return Promise.reject(new Error('Password must be text.'));
+      throw new Error('Password must be text.');
     }
 
     const wordlists = Object.values(bip39.wordlists);
@@ -113,7 +113,7 @@ class KeyringController extends EventEmitter {
         (wordlist) => !bip39.validateMnemonic(seedPhraseAsBuffer, wordlist),
       )
     ) {
-      return Promise.reject(new Error('Seed phrase is invalid.'));
+      throw new Error('Seed phrase is invalid.');
     }
 
     this.clearKeyrings();
@@ -170,7 +170,7 @@ class KeyringController extends EventEmitter {
   async submitPassword(password) {
     this.keyrings = await this.unlockKeyrings(password);
     this.setUnlocked();
-    this.fullUpdate(); // needs await?
+    this.fullUpdate();
   }
 
   /**
@@ -315,8 +315,7 @@ class KeyringController extends EventEmitter {
    */
   async exportAccount(address) {
     const keyring = await this.getKeyringForAccount(address);
-    const exported = keyring.exportAccount(normalizeAddress(address));
-    return exported;
+    return await keyring.exportAccount(normalizeAddress(address));
   }
 
   /**
@@ -370,7 +369,7 @@ class KeyringController extends EventEmitter {
   async signTransaction(ethTx, _fromAddress, opts = {}) {
     const fromAddress = normalizeAddress(_fromAddress);
     const keyring = await this.getKeyringForAccount(fromAddress);
-    return keyring.signTransaction(fromAddress, ethTx, opts);
+    return await keyring.signTransaction(fromAddress, ethTx, opts);
   }
 
   /**
@@ -384,7 +383,7 @@ class KeyringController extends EventEmitter {
   async signMessage(msgParams, opts = {}) {
     const address = normalizeAddress(msgParams.from);
     const keyring = await this.getKeyringForAccount(address);
-    return keyring.signMessage(address, msgParams.data, opts);
+    return await keyring.signMessage(address, msgParams.data, opts);
   }
 
   /**
@@ -399,7 +398,7 @@ class KeyringController extends EventEmitter {
   async signPersonalMessage(msgParams, opts = {}) {
     const address = normalizeAddress(msgParams.from);
     const keyring = await this.getKeyringForAccount(address);
-    return keyring.signPersonalMessage(address, msgParams.data, opts);
+    return await keyring.signPersonalMessage(address, msgParams.data, opts);
   }
 
   /**
@@ -413,7 +412,7 @@ class KeyringController extends EventEmitter {
   async getEncryptionPublicKey(_address, opts = {}) {
     const address = normalizeAddress(_address);
     const keyring = await this.getKeyringForAccount(address);
-    return keyring.getEncryptionPublicKey(address, opts);
+    return await keyring.getEncryptionPublicKey(address, opts);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -344,34 +329,28 @@ class KeyringController extends EventEmitter {
...
+    } else {
+      throw new Error(
+        `Keyring ${keyring.type} doesn't support account removal operations`,
+      );
+    }
+
+    const accounts = await keyring.getAccounts();
+    // Check if this was the last/only account
+    if (accounts.length === 0) {
+      await this.removeEmptyKeyrings();
```
mcmire, line 348: Looks like we might want to say:

``` suggestion
      return await this.removeEmptyKeyrings();
```

index.js
```diff
@@ -324,14 +313,10 @@ class KeyringController extends EventEmitter {
...
-      return this.getKeyringForAccount(address).then((keyring) => {
-        return keyring.exportAccount(normalizeAddress(address));
-      });
-    } catch (e) {
-      return Promise.reject(e);
-    }
+  async exportAccount(address) {
+    const keyring = await this.getKeyringForAccount(address);
+    const exported = keyring.exportAccount(normalizeAddress(address));
+    return exported;
```
mcmire, line 319: We should be able to shorten this to:

``` suggestion
    return await keyring.exportAccount(normalizeAddress(address));
```

index.js
```diff
@@ -170,12 +167,10 @@ class KeyringController extends EventEmitter {
...
-  submitPassword(password) {
-    return this.unlockKeyrings(password).then((keyrings) => {
-      this.keyrings = keyrings;
-      this.setUnlocked();
-      return this.fullUpdate();
-    });
+  async submitPassword(password) {
+    this.keyrings = await this.unlockKeyrings(password);
+    this.setUnlocked();
+    this.fullUpdate(); // needs await?
```
mcmire, line 173: `fullUpdate` doesn't seem like an `async` method, so I don't think we need to await it.

index.js
```diff
@@ -388,11 +367,10 @@ class KeyringController extends EventEmitter {
...
    * @returns {Promise<Object>} The signed transaction object.
    */
-  signTransaction(ethTx, _fromAddress, opts = {}) {
+  async signTransaction(ethTx, _fromAddress, opts = {}) {
     const fromAddress = normalizeAddress(_fromAddress);
-    return this.getKeyringForAccount(fromAddress).then((keyring) => {
-      return keyring.signTransaction(fromAddress, ethTx, opts);
-    });
+    const keyring = await this.getKeyringForAccount(fromAddress);
+    return keyring.signTransaction(fromAddress, ethTx, opts);
```
mcmire, line 373: Nit: Do we want to `await` this?

``` suggestion
    return await keyring.signTransaction(fromAddress, ethTx, opts);
```

index.js
```diff
@@ -403,11 +381,10 @@ class KeyringController extends EventEmitter {
...
    * @returns {Promise<Buffer>} The raw signature.
    */
-  signMessage(msgParams, opts = {}) {
+  async signMessage(msgParams, opts = {}) {
     const address = normalizeAddress(msgParams.from);
-    return this.getKeyringForAccount(address).then((keyring) => {
-      return keyring.signMessage(address, msgParams.data, opts);
-    });
+    const keyring = await this.getKeyringForAccount(address);
+    return keyring.signMessage(address, msgParams.data, opts);
```
mcmire, line 387: Nit: Do we want to `await` this?

``` suggestion
    return await keyring.signMessage(address, msgParams.data, opts);
```

index.js
```diff
@@ -419,11 +396,10 @@ class KeyringController extends EventEmitter {
...
    * @returns {Promise<Buffer>} The raw signature.
    */
-  signPersonalMessage(msgParams, opts = {}) {
+  async signPersonalMessage(msgParams, opts = {}) {
     const address = normalizeAddress(msgParams.from);
-    return this.getKeyringForAccount(address).then((keyring) => {
-      return keyring.signPersonalMessage(address, msgParams.data, opts);
-    });
+    const keyring = await this.getKeyringForAccount(address);
+    return keyring.signPersonalMessage(address, msgParams.data, opts);
```
mcmire, line 402: Nit: Do we want to `await` this? (and other examples)

index.js
```diff
@@ -434,11 +410,10 @@ class KeyringController extends EventEmitter {
...
    * @returns {Promise<Buffer>} The public key.
    */
-  getEncryptionPublicKey(_address, opts = {}) {
+  async getEncryptionPublicKey(_address, opts = {}) {
     const address = normalizeAddress(_address);
-    return this.getKeyringForAccount(address).then((keyring) => {
-      return keyring.getEncryptionPublicKey(address, opts);
-    });
+    const keyring = await this.getKeyringForAccount(address);
+    return keyring.getEncryptionPublicKey(address, opts);
```
mcmire, line 416: Nit: Do we want to `await` this?

mcmire (commented):
Good idea! Had some suggestions.

index.js
```diff
@@ -97,7 +97,7 @@ class KeyringController extends EventEmitter {
    * either as a string or an array of UTF-8 bytes that represent the string.
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
-  createNewVaultAndRestore(password, seedPhrase) {
+  async createNewVaultAndRestore(password, seedPhrase) {
```
Gudahtt, line 100: Nit: In an `async` function it's more idiomatic to throw errors rather than return rejections, so we should probably update the two rejections below here as well

index.js
```diff
@@ -692,35 +660,35 @@ class KeyringController extends EventEmitter {
    * @param {string} address - An account address.
```
Gudahtt, line 660: Nit: There is a `.then` a few lines up from here in `getAccounts` that we can remove as well

index.js
```diff
@@ -546,15 +517,13 @@ class KeyringController extends EventEmitter {
...
-        new Error('KeyringController - password is not a string'),
-      );
+      throw new Error('KeyringController - password is not a string');
     }
 
     this.password = password;
-    return Promise.all(
+    const serializedKeyrings = await Promise.all(
       this.keyrings.map((keyring) => {
         return Promise.all([keyring.type, keyring.serialize()]).then(
```
Gudahtt, line 528: Nit: We can replace the `.then` here with an `await` as well

Gudahtt (commented):
Great work @darkwing ! This has been on my wishlist for a long time. A few issues and suggestions but everything not commented on looks good

index.js
darkwing: Good call -- sorry for leaving this in, was a TODO for me and I just forgot to remove it.

index.js
darkwing: Good shout -- I did this elsewhere but not here.  Fixed.

index.js
darkwing: Done!

ryanml (approved):
awesome!

Gudahtt (approved):
LGTM!

PR #150, opened 2022-09-22
Update to preferred engine of Node 16

darkwing:
Node 15 has support for the WebCrypto API, which will be helpful with my other PR:

https://github.com/MetaMask/KeyringController/pull/147


Gudahtt:
We can't use the web crypto API in this package. The encryption library is injected - mobile needs to use an entirely different package.

PR #152, opened 2022-09-30
Keep User Logged In: Export key for encrypted key login

darkwing:
This is an alternative to https://github.com/MetaMask/KeyringController/pull/147 , whereby we create an encryption key based off of the derived key of browser-passworder

Depends on: https://github.com/MetaMask/browser-passworder/pull/20

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,7 +84,7 @@ class KeyringController extends EventEmitter {
 
     await this.createFirstKeyTree(password);
     await this.persistAllKeyrings();
-    this.setUnlocked.bind();
+    this.setUnlocked();
     this.fullUpdate();
   }
 
--- a/index.js
+++ b/index.js
@@ -41,8 +41,9 @@ class KeyringController extends EventEmitter {
     this.store = new ObservableStore(initState);
     this.memStore = new ObservableStore({
       isUnlocked: false,
-      keyringTypes: this.keyringTypes.map((krt) => krt.type),
+      keyringTypes: this.keyringTypes.map((keyring) => keyring.type),
       keyrings: [],
+      encryptionKey: null,
     });
 
     this.encryptor = opts.encryptor || encryptor;
@@ -150,12 +151,11 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async setLocked() {
-    delete this.encryptionKey;
     delete this.encryptionData;
     delete this.password;
 
     // set locked
-    this.memStore.updateState({ isUnlocked: false });
+    this.memStore.updateState({ isUnlocked: false, encryptionKey: null });
 
     // remove keyrings
     this.keyrings = [];
@@ -550,7 +550,8 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<boolean>} Resolves to true once keyrings are persisted.
    */
   async persistAllKeyrings() {
-    if (!this.password && !this.encryptionKey) {
+    const { encryptionKey } = this.memStore.getState();
+    if (!this.password && !encryptionKey) {
       throw new Error(
         'Cannot persist vault without password and encryption key',
       );
@@ -568,9 +569,12 @@ class KeyringController extends EventEmitter {
 
     let vault;
     if (this.password) {
-      vault = await this.encryptor.encrypt(this.password, serializedKeyrings);
-    } else if (this.encryptionKey) {
-      const key = await this.encryptor.createKeyFromString(this.encryptionKey);
+      const { vault: newVault, extractedKeyString } =
+        await this.encryptor.encrypt(this.password, serializedKeyrings);
+      vault = newVault;
+      this.memStore.updateState({ encryptionKey: extractedKeyString });
+    } else if (encryptionKey) {
+      const key = await this.encryptor.createKeyFromString(encryptionKey);
       const vaultJSON = await this.encryptor.encryptWithKey(
         key,
         serializedKeyrings,
@@ -612,16 +616,16 @@ class KeyringController extends EventEmitter {
       vault = result.vault;
 
       this.password = password;
-      this.encryptionKey = result.extractedKeyString;
       this.encryptionData = result.data;
+      this.memStore.updateState({ encryptionKey: result.extractedKeyString });
     } else {
       vault = await this.encryptor.decryptWithEncryptedKeyString(
         encryptionKey,
         encryptionData,
       );
 
-      this.encryptionKey = encryptionKey;
       this.encryptionData = encryptionData;
+      this.memStore.updateState({ encryptionKey });
     }
 
     await Promise.all(vault.map(this._restoreKeyring.bind(this)));
@@ -678,7 +682,7 @@ class KeyringController extends EventEmitter {
    * @returns {Keyring|undefined} The class, if it exists.
    */
   getKeyringClassForType(type) {
-    return this.keyringTypes.find((kr) => kr.type === type);
+    return this.keyringTypes.find((keyring) => keyring.type === type);
   }
 
   /**
--- a/index.js
+++ b/index.js
@@ -150,14 +150,13 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async setLocked() {
-    
     delete this.encryptionKey;
     delete this.encryptionData;
     delete this.password;
 
     // set locked
     this.memStore.updateState({ isUnlocked: false });
-    
+
     // remove keyrings
     this.keyrings = [];
     await this._updateMemStoreKeyrings();
@@ -181,11 +180,11 @@ class KeyringController extends EventEmitter {
   async submitPassword(password) {
     // If a user is trying to submit a password, we should lock
     // in the event that the password is incorrect
-    this.setLocked()
+    this.setLocked();
 
     await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
-    
+
     await this.persistAllKeyrings();
 
     this.setUnlocked();
@@ -570,7 +569,6 @@ class KeyringController extends EventEmitter {
     let vault;
     if (this.password) {
       vault = await this.encryptor.encrypt(this.password, serializedKeyrings);
-
     } else if (this.encryptionKey) {
       const key = await this.encryptor.createKeyFromString(this.encryptionKey);
       const vaultJSON = await this.encryptor.encryptWithKey(
@@ -579,7 +577,6 @@ class KeyringController extends EventEmitter {
       );
       vaultJSON.salt = JSON.parse(this.encryptionData).salt;
       vault = JSON.stringify(vaultJSON);
-
     }
 
     if (!vault) {
@@ -611,6 +608,7 @@ class KeyringController extends EventEmitter {
     let vault;
     if (password) {
       const result = await this.encryptor.decrypt(password, encryptedVault);
+
       vault = result.vault;
 
       this.password = password;
@@ -622,15 +620,9 @@ class KeyringController extends EventEmitter {
         encryptionData,
       );
 
-      if(!vault) {
-        console.log('[unlockKeyrings] no vault via encrypted key string: ', encryptionKey);
-      }
-
       this.encryptionKey = encryptionKey;
       this.encryptionData = encryptionData;
     }
-
-    // console.log('vault is: ', vault);
 
     await Promise.all(vault.map(this._restoreKeyring.bind(this)));
     await this._updateMemStoreKeyrings();
--- a/index.js
+++ b/index.js
@@ -842,7 +842,6 @@ class KeyringController extends EventEmitter {
     if (keyring.forgetDevice) {
       keyring.forgetDevice();
       this.persistAllKeyrings();
-      this._updateMemStoreKeyrings.bind(this)();
     } else {
       throw new Error(
         `KeyringController - keyring does not have method "forgetDevice", keyring type: ${keyring.type}`,
--- a/index.js
+++ b/index.js
@@ -573,7 +573,10 @@ class KeyringController extends EventEmitter {
 
     if (this.password) {
       const { vault: newVault, extractedKeyString } =
-        await this.encryptor.encrypt(this.password, serializedKeyrings);
+        await this.encryptor.encryptWithDetail(
+          this.password,
+          serializedKeyrings,
+        );
 
       vault = newVault;
       newExtractedKeyString = extractedKeyString;
@@ -621,7 +624,10 @@ class KeyringController extends EventEmitter {
 
     let vault;
     if (password) {
-      const result = await this.encryptor.decrypt(password, encryptedVault);
+      const result = await this.encryptor.decryptWithDetail(
+        password,
+        encryptedVault,
+      );
       vault = result.vault;
       this.encryptionSalt = result.salt;
       this.password = password;
--- a/index.js
+++ b/index.js
@@ -128,7 +128,6 @@ class KeyringController extends EventEmitter {
 
     this.clearKeyrings();
 
-    await this.persistAllKeyrings();
     const firstKeyring = await this.addNewKeyring(
       KEYRINGS_TYPE_MAP.HD_KEYRING,
       {
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ class KeyringController extends EventEmitter {
     this.store = new ObservableStore(initState);
     this.memStore = new ObservableStore({
       isUnlocked: false,
-      keyringTypes: this.keyringTypes.map((keyring) => keyring.type),
+      keyringTypes: this.keyringTypes.map((keyringType) => keyringType.type),
       keyrings: [],
       encryptionKey: null,
     });
@@ -191,6 +191,17 @@ class KeyringController extends EventEmitter {
     this.fullUpdate();
   }
 
+  /**
+   * Submit Encryption Key
+   *
+   * Attempts to decrypt the current vault and load its keyrings
+   * into memory based on the vault and CryptoKey information
+   *
+   * @emits KeyringController#unlock
+   * @param {string} encryptionKey - The encrypted key information used to decrypt the vault
+   * @param {string} loginData - The vault data to decrypt
+   * @returns {Promise<Object>} A Promise that resolves to the state.
+   */
   async submitEncryptionKey(encryptionKey, loginData) {
     this.keyrings = await this.unlockKeyrings(
       undefined,
@@ -551,6 +562,7 @@ class KeyringController extends EventEmitter {
    */
   async persistAllKeyrings() {
     const { encryptionKey } = this.memStore.getState();
+
     if (!this.password && !encryptionKey) {
       throw new Error(
         'Cannot persist vault without password and encryption key',
--- a/index.js
+++ b/index.js
@@ -185,6 +185,8 @@ class KeyringController extends EventEmitter {
     await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
 
+    this.password = password;
+
     await this.persistAllKeyrings();
 
     this.setUnlocked();
--- a/index.js
+++ b/index.js
@@ -187,8 +187,6 @@ class KeyringController extends EventEmitter {
 
     this.password = password;
 
-    await this.persistAllKeyrings();
-
     this.setUnlocked();
     this.fullUpdate();
   }
@@ -632,11 +630,10 @@ class KeyringController extends EventEmitter {
     let vault;
     if (password) {
       const result = await this.encryptor.decrypt(password, encryptedVault);
-
       vault = result.vault;
-
-      this.password = password;
       this.encryptionSalt = result.salt;
+      this.password = password;
+
       this.memStore.updateState({ encryptionKey: result.extractedKeyString });
     } else {
       vault = await this.encryptor.decryptWithEncryptedKeyString(
--- a/index.js
+++ b/index.js
@@ -151,7 +151,7 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async setLocked() {
-    delete this.encryptionData;
+    delete this.encryptionSalt;
     delete this.password;
 
     // set locked
@@ -591,7 +591,7 @@ class KeyringController extends EventEmitter {
         key,
         serializedKeyrings,
       );
-      vaultJSON.salt = JSON.parse(this.encryptionData).salt;
+      vaultJSON.salt = this.encryptionSalt;
       vault = JSON.stringify(vaultJSON);
     }
 
@@ -628,7 +628,7 @@ class KeyringController extends EventEmitter {
       vault = result.vault;
 
       this.password = password;
-      this.encryptionData = result.data;
+      this.encryptionSalt = result.salt;
       this.memStore.updateState({ encryptionKey: result.extractedKeyString });
     } else {
       vault = await this.encryptor.decryptWithEncryptedKeyString(
@@ -636,7 +636,7 @@ class KeyringController extends EventEmitter {
         encryptionData,
       );
 
-      this.encryptionData = encryptionData;
+      this.encryptionSalt = JSON.parse(encryptionData).salt;
       this.memStore.updateState({ encryptionKey });
     }
 
--- a/index.js
+++ b/index.js
@@ -176,7 +176,6 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async submitPassword(password) {
-    await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
 
     this.setUnlocked();
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ const KEYRINGS_TYPE_MAP = {
   HD_KEYRING: 'HD Key Tree',
   SIMPLE_KEYRING: 'Simple Key Pair',
 };
+
 /**
  * Strip the hex prefix from an address, if present
  * @param {string} address - The address that might be hex prefixed.
@@ -145,7 +146,7 @@ class KeyringController extends EventEmitter {
    */
   async setLocked() {
     // set locked
-    this.password = null;
+    delete this.encryptionKey;
     this.memStore.updateState({ isUnlocked: false });
     // remove keyrings
     this.keyrings = [];
@@ -168,7 +169,15 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async submitPassword(password) {
+    await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
+    await this.persistAllKeyrings(password);
+    this.setUnlocked();
+    this.fullUpdate();
+  }
+
+  async submitEncryptionKey(encryptionKey) {
+    this.keyrings = await this.unlockKeyrings(undefined, encryptionKey);
     this.setUnlocked();
     this.fullUpdate();
   }
@@ -202,7 +211,7 @@ class KeyringController extends EventEmitter {
    * @param {Object} opts - The constructor options for the keyring.
    * @returns {Promise<Keyring>} The new keyring.
    */
-  async addNewKeyring(type, opts) {
+  async addNewKeyring(type, opts, password) {
     const Keyring = this.getKeyringClassForType(type);
     const keyring = new Keyring(opts);
     if ((!opts || !opts.mnemonic) && type === KEYRINGS_TYPE_MAP.HD_KEYRING) {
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -548,15 +621,33 @@ class KeyringController extends EventEmitter {
...
+      this.encryptionSalt = result.salt;
+      this.memStore.updateState({ encryptionKey: result.extractedKeyString });
+    } else {
+      vault = await this.encryptor.decryptWithEncryptedKeyString(
+        encryptionKey,
+        encryptionData,
+      );
+
+      this.encryptionSalt = JSON.parse(encryptionData).salt;
+      this.memStore.updateState({ encryptionKey });
```
darkwing, line 648: Nit: Make sure we don't need this update if this value didn't change.

index.js
```diff
@@ -168,9 +178,42 @@ class KeyringController extends EventEmitter {
...
+    // If a user is trying to submit a password, we should lock
+    // in the event that the password is incorrect
+    this.setLocked();
+
+    await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
+
+    this.password = password;
+
+    await this.persistAllKeyrings();
```
darkwing, line 190: Nit: Do we need this call?

index.js
darkwing: Confirmed we *do* need this.

index.js
```diff
@@ -168,7 +177,36 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async submitPassword(password) {
+    // If a user is trying to submit a password, we should lock
+    // in the event that the password is incorrect
+    this.setLocked();
```
naugtur, line 182: (could) document (with a comment or in jsdoc)  the fact that setLocked can have no asynchronous side-effects. We don't want to end up running unlockKeyrings and those sideeffects in non-deterministic order.

package.json
```diff
@@ -33,7 +33,7 @@
   "dependencies": {
     "@metamask/bip39": "^4.0.0",
     "@metamask/eth-hd-keyring": "^4.0.2",
-    "browser-passworder": "^2.0.3",
+    "browser-passworder": "MetaMask/browser-passworder#export-key",
```
Gudahtt, line 36: I'd recommend using a commit hash here rather than a branch. I've found in the past that referencing a branch can result in confusing situations where an older commit is still used even after the branch is updated, because Yarn has cached the resolution of this branch to the older commit. Not 100% sure whether it still works this way though.

Gudahtt:
How will this work with mobile? Mobile's decryptor doesn't have any equivalent function to `encryptWithDetail`/`decryptWithDetail`, nor does it have any need for it.

package.json
darkwing: You're right, we do still get weird cached commit problems.  I stuck with branch as I was making frequent changes and constantly update the commit message was frustrating.

Gudahtt: Fair enough!

index.js
```diff
@@ -513,15 +546,17 @@ class KeyringController extends EventEmitter {
...
    */
-  async persistAllKeyrings(password = this.password) {
-    if (typeof password !== 'string') {
-      throw new Error('KeyringController - password is not a string');
+  async persistAllKeyrings() {
+    const { encryptionKey } = this.memStore.getState();
+
+    if (!this.password && !encryptionKey) {
+      throw new Error(
+        'Cannot persist vault without password and encryption key',
```
legobeat, line 556: Realizing I'm probably late to the party here but: How ingrained is the use of the word "password" in user-facing messages wrt this passphrase?

The reason why I'd like to be meticulous and (however late) make clear distinctions is to help users set expectations:
* A `password` is validated against some source of truth to perform authorization. There is always *some* way to reset it to regain access.
* A `passphrase` is used to derive an encryption key. If it's lost it and there are no backups, there is no recourse.

index.js
```diff
@@ -531,11 +566,41 @@ class KeyringController extends EventEmitter {
...
-      serializedKeyrings,
-    );
-    this.store.updateState({ vault: encryptedString });
+
+    let vault;
+    let newExportedKeyString;
+
+    if (this.password) {
+      const { vault: newVault, exportedKeyString } =
+        await this.encryptor.encryptWithDetail(
```
legobeat, line 575: An alternative approach (which should not require API changes to `browser-passworder`) would be to generate the key itself here (if it does not exist) and use the existing `encryptWithKeyMethod`. That would reduce the code complexity and reduce the number of potential states.

Was this considered, and if so, why is it not preferred?

index.js
```diff
@@ -168,7 +177,35 @@ class KeyringController extends EventEmitter {
...
    */
   async submitPassword(password) {
+    // If a user is trying to submit a password, we should lock
+    // in the event that the password is incorrect
+    this.setLocked();
+
+    await this.verifyPassword(password);
     this.keyrings = await this.unlockKeyrings(password);
+
+    this.password = password;
```
legobeat, line 187: While the passphrase is already set as a private variable (so shouldn't be any new exposure introduced by this PR), this sets of bells for me.

Could we get away without keeping it in memory (since the decrypted key seems to be as well)? If not, could we at least encapsulate it further? (cf. https://github.com/MetaMask/eslint-config/pull/244)

It seems reasonable that even if so, that can be addressed in a separate PR from this.

index.js
Gudahtt: This package has to work across both mobile and extension, and they use entirely different encryption methods. Mobile uses a native AES API, whereas extension uses the web crypto API. This is managed by keeping all cryptography in the "encryptor" instance, which is injected into this class.

Gudahtt: Could you elaborate on how that would reduce the number of potential states? I don't follow that part.

index.js
darkwing: I'm happy to update to either word, but I don't think it should block this PR.

index.js
```diff
@@ -531,11 +565,45 @@ class KeyringController extends EventEmitter {
...
+      vault = await this.encryptor.encrypt(this.password, serializedKeyrings);
+    }
+
+    if (!vault) {
+      throw new Error('Cannot persist vault without vault information');
+    }
+
+    this.store.updateState({ vault });
+
+    // The keyring updates need to be announced before updating the exportedKeyStrings
```
Gudahtt, line 601: What is the relationship between these two updates? It's not clear to me why this order is important.

Gudahtt:
The `cacheEncryptionKey` option looks good! Simpler than I thought it would be

index.js
darkwing: Grrrr.  It's a weird extension issue.  If I call `this.memStore.updateState({ encryptionKey: newExportedKeyString });` *before* `_updateMemStoreKeyrings`, the account is created in the keyring *but* somehow isn't being propagated to the extension before the Preferences controller tries to `syncAddresses`.

darkwing: The visual side effect is that the new account wouldn't be selected in the extension -- Account 1 would be selected.  When you open up the accounts menu, however, it's there.

index.js
legobeat: I was loosely referring to branching complexity. Inlining the logic in `encryptWithDetail` could reduce the call sites into `encryptor`, parallel logic, and code surface area. Consider:

```
        ___
_______/   \______________
       \___/
```

vs

```
     _____    ________
____/     \__/        \____
    \_____/  \________/
```

Conversely for decrypt.

 It's not a large number of codes we're talking about here, but when juggling passphrases and exposed private keys around it's prudent to be more picky IMO.

Gudahtt: Ah I see. I don't think it would reduce branching complexity overall, but it would shift it from that package into this one.

Either way, we will have to persist the keyring state with a password and no key, or with a key and no password. And the same goes for unlock: sometimes we'll have a password but no key, and sometimes we'll have a key but no password. That's two paths each, for unlock and persist.

That's four paths for the extension at least. Handling mobile as well is more complex, as their encryptor doesn't yet support returning a key (it has no immediate need to support this). So that's an additional path for unlock and persist (where a password is supplied, but we don't want to ask for a key and cache it).

Even if we inline some of the encryptor changes here, we would still have those 6 branches, the same as we have now.

Gudahtt: We could eliminate two of those branches by updating the mobile encryptor to also support `encryptWithDetail`. That'd make it easier for us to combine it with `encrypt` and eliminate the extra two methods on the encryptor as well.

But it's unclear to me how that would benefit mobile. I guess it would let them keep the key in-memory and let the garbage collector cleanup the password, which.... seems preferable but it's hard to argue it has any tangible benefit to security beyond allowing us to simplify some of this code. Postponing that change for later makes sense to me, especially given that a lot of things are waiting on this change (_a lot_ of MV3 bugs surrounding the restart of the service worker are linked to this work).

Gudahtt: If we just wanted to reduce the API surface of the encryptor, that would be a bit easier to accomplish. We could update `encrypt` and `decrypt` to return an object, and add the details to that rather than adding the new `encryptWithDetails`/`decryptWithDetails` functions.

That'd make the extension encryptor act differently from the mobile encryptor for those methods, so we'd need to handle that difference or change the mobile one as well. But that wouldn't be too difficult either way. I expect we'll want to do this eventually.

index.js
```diff
@@ -131,7 +141,7 @@ class KeyringController extends EventEmitter {
       throw new Error('KeyringController - First Account not found.');
     }
 
-    await this.persistAllKeyrings(password);
+    await this.persistAllKeyrings();
```
naugtur, line 144: Why does this need to be called twice? Looks like the initial call is redundant or there's not enough separation of concerns for addNewKeyring to work regardless.

index.js
```diff
@@ -513,15 +549,17 @@ class KeyringController extends EventEmitter {
...
-   * @param {string} password - The keyring controller password.
    * @returns {Promise<boolean>} Resolves to true once keyrings are persisted.
    */
-  async persistAllKeyrings(password = this.password) {
-    if (typeof password !== 'string') {
-      throw new Error('KeyringController - password is not a string');
+  async persistAllKeyrings() {
+    const { encryptionKey } = this.memStore.getState();
+
+    if (!this.password && !encryptionKey) {
```
naugtur, line 557: What's the difference between holding something in memory store vs class field? Why do we use/need both?

index.js
naugtur: This comment could be covering the why insteadof/alongwith the what

index.js
```diff
@@ -548,15 +620,44 @@ class KeyringController extends EventEmitter {
...
+
+        this.encryptionSalt = JSON.parse(encryptedVault).salt;
+        // This call is required on the first call because encryptionKey
+        // is not yet inside the memStore
+        this.memStore.updateState({ encryptionKey });
+      }
+    } else {
+      vault = await this.encryptor.decrypt(password, encryptedVault);
+      this.password = password;
+    }
```
naugtur, line 659: (nit) This if structure looks similar across both functions. I'd be tempted to create a function that accepts state and descriptively named functions as reactions to certain states.

index.js
naugtur: Also, could a test cover this to guarantee it's not undone?

naugtur (approved):
Left some comments, but none of them are blockers. 
Please consider addressing them before merging, but you have my approval already. This is Fiiiine!

package.json
```diff
@@ -33,7 +33,7 @@
   "dependencies": {
     "@metamask/bip39": "^4.0.0",
     "@metamask/eth-hd-keyring": "^4.0.2",
-    "browser-passworder": "MetaMask/browser-passworder#e66f3ca0b57f4ef9de4acd782e92b850cb0fe6bb",
+    "browser-passworder": "darkwing/browser-passworder#mv3-login-dist",
```
legobeat, line 36: Before merging this PR, those changes should ideally be merged to `MetaMask/browser-passworder`. If there is a blocker for this, let's make it a branch of the upstream repo and point at the commit hash? 

The reasoning is to not have builds produce different results depending on the current state of branch in an individual's external repo.

index.js
darkwing: Putting the `encryptionKey` in the memory store bubbles it up to the extension, which allows us to store it in session storage: https://github.com/MetaMask/metamask-extension/pull/15558/files#diff-6fbff2cfe97ac01b77296ef2122c7e0a5b3ff6a84b584b4d1a87482f35eea3d6R3941

Gudahtt: We don't generally have a use for _public_ class fields; the password field should be private. That's a change I was going to make in a separate PR at some point; it would be a breaking change because `@metamask/controllers` uses this field today, but they don't need to.

package.json
Gudahtt: This has just been published to npm as v4.0.1

index.js
```diff
@@ -78,9 +84,11 @@ class KeyringController extends EventEmitter {
...
    */
   async createNewVaultAndKeychain(password) {
-    await this.createFirstKeyTree(password);
-    await this.persistAllKeyrings(password);
-    this.setUnlocked.bind();
+    this.password = password;
+
+    await this.createFirstKeyTree();
+    await this.persistAllKeyrings();
+    this.setUnlocked();
```
Gudahtt, line 91: Hmm. This change caught my eye because it's a functional change (the previous `this.setUnlocked.bind()` wouldn't actually call `this.setUnlocked()`). But it looks like it was supposed to; this was an accidental change introduced in the async/await PR.

index.js
darkwing: Heh, it's from the first ever commit:
FirstCommit

index.js
```diff
@@ -168,7 +180,34 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async submitPassword(password) {
+    // If a user is trying to submit a password, we should lock
+    // in the event that the password is incorrect
+    this.setLocked();
```
Gudahtt, line 185: What motivated this change? We should only be calling `submitPassword` when the keyring is already locked, so this might not be entirely necessary.

I ask because this is introducing a new state update, as a side-effect of `setLocked`, and I'm not sure what effect it will have. And to make matters worse, there is an asynchronous step in `setLocked` which could cause confusing (the `lock` event might end up being emitted after unlocked, which is confusing).

darkwing: It's not something I consider absolutely necessary but allowing to submit password, while already unlocked and populated, seems weird.  Like, if you go to a login page but you're already logged in, you get redirected to the home screen.  Like I said, not strictly necessary, but seems something should happen here.

index.js
Gudahtt: > If I call `this.memStore.updateState({ encryptionKey: newExportedKeyString });` *before* `_updateMemStoreKeyrings`, the account is created in the keyring *but* somehow isn't being propagated to the extension before the Preferences controller tries to `syncAddresses`.  

Is this also why the calls to `_updateMemStoreKeyrings` were removed in a number of places? I was curious about that as well.

index.js
darkwing: I've removed this, as it isn't strictly related to this PR.

index.js
```diff
@@ -513,15 +549,17 @@ class KeyringController extends EventEmitter {
    * encrypts that array with the provided `password`,
    * and persists that encrypted string to storage.
    *
-   * @param {string} password - The keyring controller password.
    * @returns {Promise<boolean>} Resolves to true once keyrings are persisted.
    */
-  async persistAllKeyrings(password = this.password) {
-    if (typeof password !== 'string') {
-      throw new Error('KeyringController - password is not a string');
+  async persistAllKeyrings() {
```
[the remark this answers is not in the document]
Gudahtt: I have separated these changes into a new PR to make this one easier to review: https://github.com/MetaMask/KeyringController/pull/154

index.js
Gudahtt: I have moved this change to a separate PR, and added a test: https://github.com/MetaMask/KeyringController/pull/155

index.js
Gudahtt: Huh this is pretty tricky. I think I follow though.

Whenever `memStore` is updated, it'll trigger some other things to run (including `syncAddresses`). By setting the encryption key here before calling `_updateMemStoreKeyrings`, these state update get triggered with an _outdated_ set of keyrings and associated accounts.

An update with stale keyrings/accounts is a problem because it has side-effects like choosing a new selected address in some cases. Whereas a stale (i.e. unset) encryption key is not a problem at all. It just gets set in the next update.

index.js
Gudahtt: Another approach to consider would be adding a `try...catch`, and locking if we get an error. That would avoid any misleading `lock` event in the happy path

index.js
```diff
@@ -522,11 +555,48 @@ class KeyringController extends EventEmitter {
...
+      throw new Error('Cannot persist vault without vault information');
+    }
+
+    this.store.updateState({ vault });
+
+    // The keyring updates need to be announced before updating the encryptionKey
+    // so that the updated keyring gets propagated to the extension first.
+    // Not calling _updateMemStoreKeyrings results in the wrong account being selected
+    // in the extension.
+    await this._updateMemStoreKeyrings();
```
Gudahtt, line 595: There is one more instance of `_updateMemStoreKeyrings` that can be removed in `forgetKeyring` (it's right after a call to `persistAllKeyrings`, so it's redundant)

index.js
```diff
@@ -539,15 +609,44 @@ class KeyringController extends EventEmitter {
...
+
+        this.memStore.updateState({ encryptionKey: result.exportedKeyString });
+      } else {
+        const key = await this.encryptor.importKey(encryptionKey);
+        vault = await this.encryptor.decryptWithKey(
+          key,
+          JSON.parse(encryptedVault),
+        );
+
+        this.encryptionSalt = JSON.parse(encryptedVault).salt;
```
Gudahtt, line 639: Nit: We can refactor this to only parse the vault once:

```suggestion
        const parsedEncryptedVault = JSON.parse(encryptedVault);
        vault = await this.encryptor.decryptWithKey(
          key,
          parsedEncryptedVault,
        );

        this.encryptionSalt = parsedEncryptedVault.salt;
```

Probably nicer to rename `parsedEncryptedVault` to `encryptedVault`, and rename the current `encryptedVault` to `serializedEncryptedVault` (or something like that), but that was harder to do in a suggestion.

index.js
```diff
@@ -522,11 +555,48 @@ class KeyringController extends EventEmitter {
...
+
+    this.store.updateState({ vault });
+
+    // The keyring updates need to be announced before updating the encryptionKey
+    // so that the updated keyring gets propagated to the extension first.
+    // Not calling _updateMemStoreKeyrings results in the wrong account being selected
+    // in the extension.
+    await this._updateMemStoreKeyrings();
+    if (newExportedKeyString) {
+      this.memStore.updateState({ encryptionKey: newExportedKeyString });
```
Gudahtt, line 596: Hmm. Should we keep the `encryptionSalt` in memory alongside the `encryptionKey`? That would help us verify whether the `encryptionKey` matches what was used to encrypt the vault.

Every time a password is used to encrypt the vault, a new encryption key + salt is generated. Once that new encrypted vault is persisted, the previous `encryptionKey` becomes obsolete - it was generated using a different salt, so it will not work to decrypt the vault anymore.

The only scenario where this might happen is if the vault is persisted 3 lines up from here, but then the service worker dies before setting the encryption key in session storage. In that scenario, IF there was already an encryption key set previously, we'd have an invalid key in session storage and the next login attempt would fail.

However if we set the salt in session storage as well, we can compare the session salt to the vault salt to validate the key prior to using it. Then we can bump the user to the login screen in this edge case, and avoid leaving them stuck at the login screen.

index.js
darkwing: Good shout!

test/lib/mock-encryptor.js
```diff
@@ -1,19 +1,64 @@
...
 module.exports = {
   encrypt: sinon.stub().callsFake(function (_password, dataObj) {
     cacheVal = dataObj;
-    return Promise.resolve(mockHex);
+
+    return Promise.resolve(MOCK_HEX);
   }),
 
-  decrypt(_password, _text) {
+  encryptWithDetail: sinon.stub().callsFake(function (_password, dataObj) {
```
Gudahtt, line 19: It'd be nice to have test coverage for a mobile encryptor as well. Maybe we can leave this as-is, and move these modifications to a new mock encryptor module? Or set two up here, and return them as named exports rather than as the default export.

Then we can ensure things blow up properly if `cacheEncryptionKey` is enabled with an encryptor that doesn't support it.

Gudahtt: Converting this to a class might make that easier, as was done here: https://github.com/MetaMask/controllers/blob/main/tests/mocks/mockEncryptor.ts

Then we can extend the base encryptor to create the one with the browser-passworder additional methods.

test/index.js
```diff
@@ -460,4 +470,81 @@ describe('KeyringController', function () {
       );
     });
   });
+
+  describe('submitEncryptionKey', function () {
+    it('sets encryption key data upon submitPassword', async function () {
```
Gudahtt, line 475: This should probably be checking that the encryption key is set.

test/index.js
```diff
@@ -460,4 +470,81 @@ describe('KeyringController', function () {
       );
     });
   });
+
+  describe('submitEncryptionKey', function () {
```
Gudahtt, line 474: Maybe `cacheEncryptionKey` is a better description of this section? Only a few tests are about the `submitEncryptionKey` function, but they all seem to relate to this option.

test/index.js
```diff
@@ -460,4 +470,81 @@ describe('KeyringController', function () {
...
+    });
+
+    it('cleans up login artifacts upon lock', async function () {
+      await keyringController.submitPassword(password);
+
+      await keyringController.setLocked();
+
+      expect(keyringController.encryptionSalt).toBeUndefined();
+      expect(keyringController.password).toBeUndefined();
+      expect(keyringController.memStore.getState().encryptionKey).toBeNull();
```
Gudahtt, line 547: The `cacheEncryptionKey` option isn't being set here, so this assertion isn't doing what it appears to be doing. It'll be `null` here not because it was cleared, but because it was never set.

test/index.js
```diff
@@ -460,4 +470,81 @@ describe('KeyringController', function () {
...
+
+      expect(keyringController.encryptor.decryptWithKey.calledOnce).toBe(true);
+      expect(keyringController.keyrings).toHaveLength(1);
+    });
+
+    it('should not load keyrings when incorrect encryptionKey', async function () {
+      keyringController.cacheEncryptionKey = true;
+      keyringController.setLocked(true);
+
+      await expect(keyringController.submitEncryptionKey(`{}`)).rejects.toThrow(
```
Gudahtt, line 500: We could use an additional test where a _valid_ but _incorrect_ encyption key is passed in. I wouldn't call `{}` an "incorrect" key, it's not a valid key, making this more of a validation test.

test/index.js
```diff
@@ -460,4 +470,81 @@ describe('KeyringController', function () {
...
+      keyringController.cacheEncryptionKey = true;
+      const returnValue = await keyringController.encryptor.decryptWithKey();
+      const stub = sinon.stub(keyringController.encryptor, 'decryptWithKey');
+      stub.resolves(Promise.resolve(returnValue));
+
+      keyringController.store.updateState({ vault: MOCK_ENCRYPTION_DATA });
+      await keyringController.submitEncryptionKey(MOCK_ENCRYPTION_KEY);
+
+      expect(keyringController.encryptor.decryptWithKey.calledOnce).toBe(true);
+      expect(keyringController.keyrings).toHaveLength(1);
```
Gudahtt, line 493: This might be a more robust test if we locked the keyring controller first before submitting the encryption key. Otherwise this keyring state might be from the initial keyring added in the `beforeEach`. It's not clear it's a result of entering the key.

Gudahtt: That would also let you get rid of the sinon stub, as you wouldn't need to verify the encryptor was called. The state would prove the key was used.

Though... the password is still set. Hmm.  Ideally the keyring controller would be constructed per-test, so you could do setup differently here and never enter a password. But that would be a larger refactor. Maybe in the meantime you can set `keyringController` to `undefined`, and we can refactor the tests separately.

test/index.js
```diff
@@ -460,4 +470,81 @@ describe('KeyringController', function () {
...
+      expect(keyringController.keyrings).toHaveLength(0);
+    });
+
+    it('persists keyrings when actions are performed', async function () {
+      keyringController.cacheEncryptionKey = true;
+      keyringController.store.updateState({ vault: MOCK_ENCRYPTION_DATA });
+      await keyringController.submitEncryptionKey(MOCK_ENCRYPTION_KEY);
+
+      const stub = sinon.stub(keyringController, 'persistAllKeyrings');
+      stub.resolves(Promise.resolve(true));
```
Gudahtt, line 513: If you just want to see how a function was called, you don't need to stub it out. Instead use `sinon.spy`, which lets you spy on the function without changing how it behaves.

```suggestion
      const spy = sinon.spy(keyringController, 'persistAllKeyrings');
```

Gudahtt: Though... generally it's not a good idea to be using internals in tests. They can serve as an unnecessary obstacle to refactoring, and generally they're farther away from testing real use cases. If we could update the assertions to avoid using internals altogether that would be idea.

In this case, the `memStore` can be inspected to see how many accounts there are - there's no need to use a spy. And the same goes for the addition of a new keyring and the removal of an account. It's all in the memStore.

index.js
```diff
@@ -513,6 +537,14 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<boolean>} Resolves to true once keyrings are persisted.
    */
   async persistAllKeyrings() {
+    const { encryptionKey } = this.memStore.getState();
+
+    if (!this.password && !encryptionKey) {
+      throw new Error(
+        'Cannot persist vault without password and encryption key',
```
Gudahtt, line 544: It'd be great to have test coverage for this error

index.js
```diff
@@ -522,11 +554,48 @@ class KeyringController extends EventEmitter {
...
+      if (this.password) {
+        const { vault: newVault, exportedKeyString } =
+          await this.encryptor.encryptWithDetail(
+            this.password,
+            serializedKeyrings,
+          );
+
+        vault = newVault;
+        newExportedKeyString = exportedKeyString;
+      } else if (encryptionKey) {
```
Gudahtt, line 571: It looks like this branch isn't tested, because all of the tests have `this.password` set.

test/index.js
darkwing: Good shout!  Updated!

test/index.js
```diff
@@ -460,4 +470,99 @@ describe('KeyringController', function () {
...
   });
+
+  describe('cacheEncryptionKey', function () {
+    it('sets encryption key data upon submitPassword', async function () {
+      keyringController.cacheEncryptionKey = true;
+      await keyringController.submitPassword(password);
+
+      expect(keyringController.password).toBe(password);
+      expect(keyringController.memStore.getState().encryptionSalt).toBe('SALT');
+      // eslint-disable-next-line jest/no-restricted-matchers
```
Gudahtt, line 481: Nit: In these cases where w want to ensure we have a non-empty string, we can use `expect(...).toStrictEqual(expect.stringMatching('.+'))`. This is a bit more specific than the truthy check in that it ensures it's a string, and it avoids the need for this ignore comment.

test/index.js
darkwing: Good shout!

test/index.js
darkwing: Excellent, I had no idea!

index.js
darkwing: Fixed!

test/index.js
darkwing: Done!

test/index.js
Gudahtt: There is one more place you could use this (on this line :sweat_smile: )

test/index.js
```diff
@@ -460,4 +470,124 @@ describe('KeyringController', function () {
...
+      expect(keyringController.memStore.getState().encryptionSalt).toBe('SALT');
+      expect(keyringController.memStore.getState().encryptionKey).toStrictEqual(
+        expect.stringMatching('.+'),
+      );
+    });
+
+    it('unlocks the keyrings with valid information', async function () {
+      keyringController.cacheEncryptionKey = true;
+      const returnValue = await keyringController.encryptor.decryptWithKey();
+      const stub = sinon.stub(keyringController.encryptor, 'decryptWithKey');
```
Gudahtt, line 489: Nit: IMO this would be a better test with this removed. The keyrings-based assertion is enough to show this works; this part just makes the test more fragile by tying it to internals.

test/index.js
```diff
@@ -460,4 +470,124 @@ describe('KeyringController', function () {
...
+          'OUTDATED_SALT',
+        ),
+      ).rejects.toThrow('Encryption key and salt provided are expired');
+      expect(keyringController.password).toBeUndefined();
+      expect(keyringController.keyrings).toHaveLength(0);
+    });
+
+    it('persists keyrings when actions are performed', async function () {
+      await keyringController.setLocked();
+      keyringController.cacheEncryptionKey = true;
```
Gudahtt, line 536: The ESLint error can be avoided by making this the first step

Gudahtt (dismissed):
LGTM! Still a lint failure to resolve, but this all looks great and I can re-approve after that's fixed.

Gudahtt (approved):

PR #154, opened 2022-11-07
Set password sooner to avoid redundant persistance

Gudahtt:
The vault password is now always set in the method responsible for creating the vault. Previously it was set when the vault was persisted, or when the first keyring was created. This led to redundant calls to persist the keyrings, just to achieve the side-effect of setting the password.

This change eliminates those redundant calls, and a few more that had no obvious purpose.

Note that the password is also set on unlock, and that has not changed here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -78,8 +78,9 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async createNewVaultAndKeychain(password) {
-    await this.createFirstKeyTree(password);
-    await this.persistAllKeyrings(password);
+    this.password = password;
+
+    await this.createFirstKeyTree();
     this.setUnlocked();
     this.fullUpdate();
   }
@@ -116,9 +117,9 @@ class KeyringController extends EventEmitter {
       throw new Error('Seed phrase is invalid.');
     }
 
-    this.clearKeyrings();
+    this.password = password;
 
-    await this.persistAllKeyrings(password);
+    this.clearKeyrings();
     const firstKeyring = await this.addNewKeyring(
       KEYRINGS_TYPE_MAP.HD_KEYRING,
       {
@@ -130,8 +131,6 @@ class KeyringController extends EventEmitter {
     if (!firstAccount) {
       throw new Error('KeyringController - First Account not found.');
     }
-
-    await this.persistAllKeyrings(password);
     this.setUnlocked();
     return this.fullUpdate();
   }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -752,7 +743,7 @@ class KeyringController extends EventEmitter {
   forgetKeyring(keyring) {
     if (keyring.forgetDevice) {
       keyring.forgetDevice();
-      this.persistAllKeyrings.bind(this)();
```
Gudahtt, line 755: This `.bind` was redundant, as we're calling `persistAllKeyrings` on `this` anyway. So it's a no-op.

index.js
```diff
@@ -78,8 +78,9 @@ class KeyringController extends EventEmitter {
    * @returns {Promise<Object>} A Promise that resolves to the state.
    */
   async createNewVaultAndKeychain(password) {
-    await this.createFirstKeyTree(password);
-    await this.persistAllKeyrings(password);
+    this.password = password;
+
+    await this.createFirstKeyTree();
```
Gudahtt, line 83: `createFirstKeyTree` only modifies `this.keyrings` via `addNewKeyring`, which calls `persistAllKeyrings` internally after adding the keyring. That's why the `persistAllKeyrings` call was eliminated here.

index.js
```diff
@@ -116,9 +117,9 @@ class KeyringController extends EventEmitter {
       throw new Error('Seed phrase is invalid.');
     }
 
-    this.clearKeyrings();
+    this.password = password;
 
-    await this.persistAllKeyrings(password);
```
Gudahtt, line 121: I'm guessing this call was primarily meant to set the password, so it has been removed.

It did have another functional impact, in that it ensured the `clearKeyrings` operation was reflected in the vault, in case the attempt to create a new vault fails. But... that situation is not anticipated. At this point we have validated that the SRP is valid, and the password is text, so there aren't any more additional failure conditions.

Moreover; leaving an existing vault in state isn't an awful end result to encountering an unexpected error. If anything it might aid with recovery, if we do somehow end up accidentally bricking the wallet with an update.

index.js
```diff
@@ -130,8 +131,6 @@ class KeyringController extends EventEmitter {
     if (!firstAccount) {
       throw new Error('KeyringController - First Account not found.');
     }
-
-    await this.persistAllKeyrings(password);
```
Gudahtt, line 134: Removed because the last time `this.keyrings` is modified is within `addNewKeyring`, which calls `persistAllKeyrings` internally.

darkwing (approved):
This makes sense!

PR #155, opened 2022-11-07
Ensure newly created vaults are unlocked

Gudahtt:
A new vault should be unlocked after being created, because the password is submitted as part of creating the vault. This was accidentally broken as part of #148. A unit test has been added to ensure this doesn't happen again.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -85,7 +85,7 @@ describe('KeyringController', function () {
   });
 
   describe('createNewVaultAndKeychain', function () {
-    it('should set a vault on the configManager', async function () {
+    it('should create a new vault', async function () {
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -85,7 +85,7 @@ describe('KeyringController', function () {
   });
 
   describe('createNewVaultAndKeychain', function () {
-    it('should set a vault on the configManager', async function () {
+    it('should create a new vault', async function () {
```
Gudahtt, line 88: This change is unrelated; I just happened to notice a reference to a now-extinct "configManager" abstraction in another test. This was the last remaining reference to it.

darkwing (approved):

PR #157, opened 2022-11-08
Migrate to Yarn v3

Gudahtt:
The project has been migrated to Yarn v3. The changes made have been based off of the module template. `allow-scripts` has been updated to the latest version as well.


README.md
```diff
@@ -50,11 +50,10 @@ Currently the methods are heavily commented in [the source code](./index.js), so
 
 ### Setup
 
-- Install [Node.js](https://nodejs.org) version 12
+- Install [Node.js](https://nodejs.org) version 14
```
Gudahtt, line 53: This change was unrelated; I just happened to notice that it was wrong. We forgot this when the Node.js minimum version was bumped to v14.

package.json
```diff
@@ -62,8 +61,11 @@
   "lavamoat": {
     "allowScripts": {
       "@lavamoat/preinstall-always-fail": false,
-      "keccak": true,
-      "secp256k1": true
+      "eth-sig-util>ethereumjs-util>keccak": false,
```
Gudahtt, line 64: These were all `true` before, but seemingly they aren't needed at all for local development, so I left them disabled.

PR #158, opened 2022-11-08
Update GitHub actions to match module template

Gudahtt:
The GitHub actions have been updated to match the module template more closely. There are still many differences because this library is not yet using TypeScript, so it doesn't have any build step. It also doesn't yet have API documentation, so steps relating to that have been removed.


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,119 @@
...
+jobs:
+  prepare:
+    name: Prepare
+    runs-on: ubuntu-latest
+    outputs:
+      YARN_CACHE_DIR: ${{ steps.yarn-cache-dir.outputs.YARN_CACHE_DIR }}
+      YARN_VERSION: ${{ steps.yarn-version.outputs.YARN_VERSION }}
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x, 18.12.1, 19.x]
```
Gudahtt, line 17: Node.js v18 has been temporarily pinned to v18.12.1 because there is a range of versions (including v18.12.0, which is the default choice of `actions/setup-node` for v18.x) that blows up during the unit tests due to an SSL error. This was fixed in v18.12.1.

We can change this back to v18.x once `actions/setup-node` updates and their caches are populated with some version >=v18.12.1

Gudahtt:
See here for the module template workflows: https://github.com/MetaMask/metamask-module-template/tree/main/.github/workflows

mcmire (approved):

PR #161, opened 2022-11-10
Return fullUpdate from methods it used to

darkwing:
After seeing a test failure in the extension, I noticed that `createNewVaultAndKeychain` wasn't returning the vault like it used to before https://github.com/MetaMask/KeyringController/commit/e797bdf51b651dd8308e4535fdcb5896f1025073.  This fixes the issue.


Gudahtt (approved):
LGTM!

PR #163, opened 2022-11-15
Adds async init method to keyrings

bergarces:
Some hardware keyrings will require an initialization step that is asynchronous and shouldn't be placed in the constructor. This PR addresses the necessary changes to call that initialization method (`init`) when required.

It also adds a missing await in the `addAccounts` method from the `HD_KEYRING`, which is asynchronous and it is currently not being awaited.


.eslintrc.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
   root: true,
+  parserOptions: {
+    ecmaVersion: 2018, // to support object rest spread, e.g. {...x, ...y}
```
bergarces, line 4: I copied this from eth-trezor-keyring in order to enable the spread operator. I don't think it should add any compatibility issues.

index.js
```diff
@@ -237,11 +240,11 @@ class KeyringController extends EventEmitter {
...
    */
   async addNewKeyring(type, opts) {
-    const Keyring = this.getKeyringClassForType(type);
-    const keyring = new Keyring(opts);
+    const keyring = await this._newKeyring(type, opts);
+
     if ((!opts || !opts.mnemonic) && type === KEYRINGS_TYPE_MAP.HD_KEYRING) {
       keyring.generateRandomMnemonic();
-      keyring.addAccounts();
+      await keyring.addAccounts();
```
bergarces, line 247: This method is asynchronous. I've taken the liberty to add an `await`.

index.js
```diff
@@ -873,6 +875,36 @@ class KeyringController extends EventEmitter {
...
+    if (keyring.init) {
+      await keyring.init();
+    }
+
+    await keyring.deserialize(data);
+
+    return keyring;
+  }
+}
+
```
bergarces, line 897: There are many instances in which we try to access the `type` from the static class. "Luckily", javascript allows a function to also have properties.

By exporting this helper method we can make sure that whenever we create a keyring builder function, the type is added as well.

index.js
```diff
@@ -873,6 +875,36 @@ class KeyringController extends EventEmitter {
       );
     }
   }
+
```
bergarces, line 878: There are two methods in which a new keyring is instantiated: `_restoreKeyring` and `addNewKeyring`.

I've extracted the logic so that we always instantiate, initialise (if available) and deserialize.

On the topic of deserialization, after looking at every single keyring I'm convinced that we don't have to pass the data at construction time as long as we call `deserialize` afterwards. `deserialize` is an asynchronous function, and we probably should stop calling it within the constructor.

index.js
bergarces: Also, since we are always calling `deserialize` after instantiating the keyring, I'm just passing the bridge to the constructor.

`deserialize` always seems return a promise, so it shouldn't be called in the constructor.

index.js
```diff
@@ -35,13 +38,15 @@ class KeyringController extends EventEmitter {
   constructor(opts) {
     super();
     const initState = opts.initState || {};
-    this.keyringTypes = opts.keyringTypes
-      ? keyringTypes.concat(opts.keyringTypes)
-      : keyringTypes;
+    this.keyringBuilders = opts.keyringBuilders
+      ? defaultKeyringBuilders.concat(opts.keyringBuilders)
+      : defaultKeyringBuilders;
```
bergarces, line 43: I thought it would be wise to rename this property as it will no longer hold class types, but builder functions.

I have checked within the extension code and it does not appear to be used, but worth thinking about any possible side effects as this can be a breaking change if external code tries to access this property.

index.js
```diff
@@ -873,6 +879,36 @@ class KeyringController extends EventEmitter {
...
+    return new KeyringClass(constructorDependencies);
+  };
+
+  builder.type = KeyringClass.type;
+
+  return builder;
 }
 
 module.exports = KeyringController;
+module.exports.keyringBuilderFactory = keyringBuilderFactory;
```
bergarces, line 914: Any strong opinions on whether we should continue exporting KeyringController as default or should instead export `KeyringController` and `keyringBuilderFactory` individually? There are already other breaking changes, so this one wouldn't be too bad.

Gudahtt: Moving `KeyringController` to a named exports sounds like a good idea to me

bergarces: I've made it a named export now.

index.js
Gudahtt: Nit: could we move this to a separate PR? We could merge this in isolation, since it's not directly tied to the other changes. That would make it easier to understand the commit history and prepare the changelog and so on.

index.js
```diff
@@ -873,6 +879,38 @@ class KeyringController extends EventEmitter {
       );
     }
   }
+
+  async _newKeyring(type, data) {
```
Gudahtt, line 883: Nit: Could you add a JSDoc comment for this function?

Gudahtt: Also for `keyringBuilderFactory`

index.js
Gudahtt: If we're changing the "Keyring" API in a non-backwards-compatible way, I'd prefer that we start using an object for constructor options rather than parameters (i.e. the "options bag" pattern, passing in `{ bridge }` rather than `bridge`). This pattern is easier to read, and makes it easy to add additional options without breaking changes.

index.js
```diff
@@ -873,6 +879,38 @@ class KeyringController extends EventEmitter {
...
+    }
+
+    return keyring;
+  }
 }
 
-module.exports = KeyringController;
+function keyringBuilderFactory(KeyringClass, BridgeClass) {
+  const builder = () => {
+    const constructorDependencies = BridgeClass ? new BridgeClass() : undefined;
```
Gudahtt, line 904: I'm torn on whether this is a good idea. I see the advantage of including a "builder" function here directly, but maybe it would be better to encourage using custom builder functions instead.

Most bridges will require initialization, for example, but this offers no way to do that.

Maybe we could instead offer a simpler "builder" function just for the simple keyrings, and recommend a custom builder function for keyrings using bridges?

index.js
Gudahtt: It would also be great to have a test for this, to demonstrate that when `addAccounts` is doing asynchronous work, this function guarantees that the accounts have been added by the time it resolves.

index.js
bergarces: Done

index.js
bergarces: I was at odds with doing this as well, probably best to leave a simple one as you suggest though.

What tilted me towards creating this method was the odd situation in which we have to make sure that the builder function returned must have a `type` property, which is not something particularly easy to document.

I've replaced it with the simpler version, so the one with the bridge will probably have to be added to `metamask-controller` in order to instantiate some of the keyrings (the QR one probably won't need a bridge and can use this function too).

index.js
bergarces: Fair enough. I've made the change to the options bag pattern directly in `eth-trezor-keyring` PR. There's no longer any reference here as the new simple builder factory does not concern itself with bridges anymore.

Gudahtt (approved):
LGTM!

index.js
bergarces: Yes, better on a different PR. I've removed it from here and will open another one.

Gudahtt:
Thanks @bergarces, that would be great! If we can get this into the extension and update the keyrings to use this builder API one-at-a-time that would be ideal.

PR #169, opened 2022-11-25
Fix: saving serialized keyring for which corresponding keyring class is not present

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/16674

For MV3 currently we are not supporting hardware wallets, this can result in error being thrown if Keyring classes of hardware wallets is not passed to Keyring Controller.

PR fixes the error by adding `null` check. Keyring for which corresponding keyring class is not present is saved in state and then re-serialized.

### DIFF
--- a/index.js
+++ b/index.js
@@ -563,7 +563,9 @@ class KeyringController extends EventEmitter {
       }),
     );
 
-    this.unsupported_keyrings.forEach(keyring => serializedKeyrings.push(keyring));
+    this.unsupported_keyrings.forEach((keyring) =>
+      serializedKeyrings.push(keyring),
+    );
 
     let vault;
     let newEncryptionKey;
@@ -702,8 +704,8 @@ class KeyringController extends EventEmitter {
 
     const Keyring = this.getKeyringClassForType(type);
     if (!Keyring) {
-      unsupported_keyrings.push(serialized);
-      return;
+      this.unsupported_keyrings.push(serialized);
+      return undefined;
     }
     const keyring = new Keyring();
     await keyring.deserialize(data);
--- a/index.js
+++ b/index.js
@@ -43,12 +43,12 @@ class KeyringController extends EventEmitter {
       isUnlocked: false,
       keyringTypes: this.keyringTypes.map((keyringType) => keyringType.type),
       keyrings: [],
-      unsupported_keyrings: [],
       encryptionKey: null,
     });
 
     this.encryptor = opts.encryptor || encryptor;
     this.keyrings = [];
+    this.unsupported_keyrings = [];
 
     // This option allows the controller to cache an exported key
     // for use in decrypting and encrypting data without password
--- a/index.js
+++ b/index.js
@@ -48,7 +48,7 @@ class KeyringController extends EventEmitter {
 
     this.encryptor = opts.encryptor || encryptor;
     this.keyrings = [];
-    this.unsupported_keyrings = [];
+    this.unsupportedKeyrings = [];
 
     // This option allows the controller to cache an exported key
     // for use in decrypting and encrypting data without password
@@ -563,7 +563,7 @@ class KeyringController extends EventEmitter {
       }),
     );
 
-    serializedKeyrings.push(...this.unsupported_keyrings);
+    serializedKeyrings.push(...this.unsupportedKeyrings);
 
     let vault;
     let newEncryptionKey;
@@ -695,14 +695,14 @@ class KeyringController extends EventEmitter {
    * On success, returns the resulting keyring instance.
    *
    * @param {Object} serialized - The serialized keyring.
-   * @returns {Promise<Keyring>} The deserialized keyring.
+   * @returns {Promise<Keyring|undefined>} The deserialized keyring or undefined if the keyring type is unsupported.
    */
   async _restoreKeyring(serialized) {
     const { type, data } = serialized;
 
     const Keyring = this.getKeyringClassForType(type);
     if (!Keyring) {
-      this.unsupported_keyrings.push(serialized);
+      this.unsupportedKeyrings.push(serialized);
       return undefined;
     }
     const keyring = new Keyring();
--- a/index.js
+++ b/index.js
@@ -563,9 +563,7 @@ class KeyringController extends EventEmitter {
       }),
     );
 
-    this.unsupported_keyrings.forEach((keyring) =>
-      serializedKeyrings.push(keyring),
-    );
+    serializedKeyrings.push(...this.unsupported_keyrings);
 
     let vault;
     let newEncryptionKey;
--- a/index.js
+++ b/index.js
@@ -43,6 +43,7 @@ class KeyringController extends EventEmitter {
       isUnlocked: false,
       keyringTypes: this.keyringTypes.map((keyringType) => keyringType.type),
       keyrings: [],
+      unsupported_keyrings: [],
       encryptionKey: null,
     });
 
@@ -562,6 +563,8 @@ class KeyringController extends EventEmitter {
       }),
     );
 
+    this.unsupported_keyrings.forEach(keyring => serializedKeyrings.push(keyring));
+
     let vault;
     let newEncryptionKey;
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -48,6 +48,7 @@ class KeyringController extends EventEmitter {
 
     this.encryptor = opts.encryptor || encryptor;
     this.keyrings = [];
+    this.unsupported_keyrings = [];
```
mcmire, line 51: Should we use `camelCase` for this property name?

index.js
```diff
@@ -562,6 +563,10 @@ class KeyringController extends EventEmitter {
       }),
     );
 
+    this.unsupported_keyrings.forEach((keyring) =>
+      serializedKeyrings.push(keyring),
+    );
```
mcmire, line 568: Nit: Can also be written

``` suggestion
    serializedKeyrings.push(...this.unsupported_keyrings);
```

index.js
```diff
@@ -679,7 +684,9 @@ class KeyringController extends EventEmitter {
    */
   async restoreKeyring(serialized) {
     const keyring = await this._restoreKeyring(serialized);
-    await this._updateMemStoreKeyrings();
+    if (keyring) {
```
mcmire, line 687: Should we add a test for `restoreKeyring` which tests this behavior?

test/index.js
```diff
@@ -94,6 +94,28 @@ describe('KeyringController', function () {
     });
   });
 
+  describe('persistAllKeyrings', function () {
+    it('should persist keyrings in unsupported_keyrings array', async function () {
```
mcmire, line 98: Why are we putting all of the keyrings in `unsupported_keyrings` when `persistAllKeyrings` is called? It makes sense that `unsupported_keyrings` would hold all keyrings that we can't find a Keyring class for, but that doesn't seem to be what this is doing. So I'm a bit confused as to how `unsupported_keyrings` would end up getting used.

test/index.js
```diff
@@ -330,6 +352,15 @@ describe('KeyringController', function () {
...
       });
     });
+    it('add serialized keyring to unsupported_keyrings array if keyring type is not known', async function () {
+      mockEncryptor.encrypt(password, [
+        { type: 'Ledger Keyring', data: 'DUMMY' },
+      ]);
+      await keyringController.setLocked();
+      const keyrings = await keyringController.unlockKeyrings(password);
+      expect(keyrings).toHaveLength(0);
+      expect(keyringController.unsupported_keyrings).toHaveLength(1);
```
mcmire, line 362: This test would pass if the code added anything to `unsupported_keyrings`. To be more accurate, what do you think about:

``` suggestion
      expect(keyringController.unsupported_keyrings).toStrictEqual(keyrings);
```

mcmire (commented):
Hey @jpuri it's possible I'm missing some context but I had a question about how the new property being added here would be used.

index.js
```diff
@@ -696,6 +703,10 @@ class KeyringController extends EventEmitter {
     const { type, data } = serialized;
 
     const Keyring = this.getKeyringClassForType(type);
+    if (!Keyring) {
+      this.unsupported_keyrings.push(serialized);
+      return undefined;
```
digiwand, line 708: we can add the new return value to line 700 above. maybe

```
* @returns {Promise<Keyring>|undefined} The deserialized keyring or undefined if the keyring type is unsupported.
```

test/index.js
Gudahtt: `persistAllKeyrings` is reading from `unsupported_keyrings`, not writing to it. We're persisting the unsupported keyrings so that they remain in the users vault, in case they need to manually extract them, or in case we add back support for that keyring.

I can see how this description could be read either way. It doesn't mean the keyrings are being persisted _to_ that array, just that that array is included in the set of data being persisted.

index.js
Gudahtt: Good catch! Slight correction, the type would be `Promise<Keyring|undefined>` since `undefined` here is still returned as a Promise.

index.js
jpuri: Done

index.js
jpuri: I have tested this [here](https://github.com/MetaMask/KeyringController/pull/169/files#diff-5bb8db779819ddef5956a5d9d5949c05ef7445237656ca37bf2f02720271440bR360) by calling `unlockKeyrings` method.

I could not find use of `restoreKeyring` method anywhere, is it added for testing purpose ?

index.js
digiwand: ah I missed the async. Thanks @Gudahtt!

index.js
Gudahtt: Good question. This method dates back long enough that I don't know the reason. I can't think of a use for it.

index.js
```diff
@@ -48,6 +48,7 @@ class KeyringController extends EventEmitter {
 
     this.encryptor = opts.encryptor || encryptor;
     this.keyrings = [];
+    this.unsupportedKeyrings = [];
```
Gudahtt, line 51: Nit: We should probably prefix this with an underscore, to denote that it's not meant to be accessed directly.

I had meant to hide the other properties as well, during the TypeScript migration. I guess this could wait until then too.

test/index.js
```diff
@@ -94,6 +94,28 @@ describe('KeyringController', function () {
     });
   });
 
+  describe('persistAllKeyrings', function () {
+    it('should persist keyrings in unsupportedKeyrings array', async function () {
+      const unsupportedKeyring = 'DUMMY_KEYRING';
+      keyringController.unsupportedKeyrings = [unsupportedKeyring];
```
Gudahtt, line 100: Nit: It would be better for this test to set the keyrings in the same way the extension would. We wouldn't ever use the keyring controller this way, setting the `unsupportedKeyrings` property directly.

Unfortunately the way these tests are setup makes that difficult. The extension passes in existing vaults via the constructor, but these tests all construct the controller the same way in `beforeEach`. This is why we've been recommending that people avoid using `beforeEach`, and use setup functions instead. Then we can create new setup functions tailored for specific tests. Improving this test in the way I'm suggesting here might require a similar refactor of the other tests first.

test/index.js
```diff
@@ -94,6 +94,28 @@ describe('KeyringController', function () {
...
+      await keyringController.persistAllKeyrings();
```
Gudahtt, line 101: Nit: typically for tests, we use newlines to separate the different phases of the test: arrange, act, and assert. Arrange is the setup, act is where we do the thing that is under test, and assert is where we check that it works.

Adding newlines to separate these sections would make this easier to read:

```suggestion

      await keyringController.persistAllKeyrings();
  
```

Likewise for the other two tests

Gudahtt (approved):
LGTM!

I agree with @mcmire that we should have a test for `restoreKeyring`, and I've left a few more suggestions on improving the tests. But I didn't want to block on those; the tests here already aren't in great shape. The basic functionality is at least covered.

segun (dismissed):

test/index.js
```diff
@@ -94,6 +94,28 @@ describe('KeyringController', function () {
...
+
+    it('emits "unlock" event', async function () {
+      await keyringController.setLocked();
+
+      const spy = sinon.spy();
+      keyringController.on('unlock', spy);
+
+      await keyringController.submitPassword(password);
+      expect(spy.calledOnce).toBe(true);
+    });
```
digiwand, line 118:
```suggestion
```

I believe we can remove this since it seems to be a duplicate of the test in `describe('submitPassword'`

index.js
jpuri: I added test case for `restoreKeyring`

index.js
jpuri: Good point, I added underscore to it.

Gudahtt (approved):

index.js
```diff
@@ -690,12 +695,16 @@ class KeyringController extends EventEmitter {
...
-   * @returns {Promise<Keyring>} The deserialized keyring.
+   * @returns {Promise<Keyring|undefined>} The deserialized keyring or undefined if the keyring type is unsupported.
    */
   async _restoreKeyring(serialized) {
     const { type, data } = serialized;
 
     const Keyring = this.getKeyringClassForType(type);
+    if (!Keyring) {
+      this._unsupportedKeyrings.push(serialized);
+      return undefined;
```
digiwand, line 706: hi @jpuri, late question: when we support hard wallets for MV3, will this logic added in this PR still be needed?

PR #176, opened 2022-12-13
await for addAccounts for HD Keyring

bergarces:
Adds missing `await` to an asynchronous call that adds accounts to a new HD Keyring.

It also includes a unit test to validate that we are awaiting for that call.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -282,54 +282,34 @@ describe('KeyringController', function () {
     });
 
     it('should finish adding HD Keyring accounts before continuing execution', async function () {
-      let timeoutSet = false;
       let addAccountsResolved = false;
 
       const originalAccAccounts = HdKeyring.prototype.addAccounts;
       sinon.stub(HdKeyring.prototype, 'addAccounts').callsFake(function () {
         return new Promise((resolve) => {
+          // By using setImmediate, this won't be resolved before getAccounts unless it is awaited
           originalAccAccounts
             .bind(this)()
             .then((accounts) => {
-              setTimeout(() => {
+              setImmediate(() => {
                 addAccountsResolved = true;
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -280,6 +280,39 @@ describe('KeyringController', function () {
 
       sinon.assert.calledOnce(initSpy);
     });
+
```
bergarces, line 283: @Gudahtt I'm not entirely sure what sort of test we should be expecting to validate what is mentioned in this comment https://github.com/MetaMask/KeyringController/pull/163#discussion_r1042417387

This test would only check that `await keyring.addAccounts()` will indeed finish its execution before the next statement is called.

I'm not sure it's particularly relevant, as we would only be testing that `await` works, which I'd say it is a given.

Gudahtt: Ideally we'd have a test that relied just on the external expectations. In this case, we expect that a keyring can have an async `addAccounts` method, and that after `addNewKeyring` is called that an initial account has been added.

Your test is already effectively using a keyring with an async `addAccounts` method, but the assertion you're using relies upon test internals rather than the external expectation a user of this method might have. That could make this as likely to break from a refactor as it would from a legitimate breakage. For example, if the "duplicate account" check was removed, we wouldn't need the `getAccounts` call anymore so this would stop working.

Instead perhaps you could check the number of accounts after `addNewKeyring` is called, to ensure the account was added correctly. Then your `addAccounts` mock could be simplified as well; it would only need to make the call async and non-eager (i.e. put a delay before the call to the original function).

Gudahtt: Thanks!

Gudahtt (approved):
LGTM!

PR #177, opened 2022-12-13
bump `@metamask/eth-hd-keyring` to v5

adonesky1:
Bumps `@metamask/eth-hd-keyring` to v5 and modifies tests accordingly

There are [a number of breaking changes](https://github.com/MetaMask/eth-hd-keyring/blob/main/CHANGELOG.md#500) associated with this version bump, but the `KeyringController` is already adapted to all such changes with the exception of expecting a public `wallets` property on the HDKeyring in a couple of tests. Those instances have been modified accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "dependencies": {
     "@metamask/bip39": "^4.0.0",
     "@metamask/browser-passworder": "^4.0.1",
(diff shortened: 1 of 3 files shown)


Gudahtt (approved):
LGTM!

Gudahtt (approved):

PR #179, opened 2022-12-14
Remove bip39 dependency

adonesky1:
Remove `bip39` dependency. The [validation removed here](https://github.com/MetaMask/KeyringController/blob/main/index.js#L127) is now taken care of in `@metamask/eth-hd-keyring`: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L284

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,37 +106,28 @@ class KeyringController extends EventEmitter {
    * creates a new encrypted store with the given password,
    * creates a new HD wallet from the given seed with 1 account.
    *
-   * @emits KeyringController#unlock
-   * @param {string} password - The password to encrypt the vault with
+   * @fires KeyringController#unlock
+   * @param {string} password - The password to encrypt the vault with.
    * @param {Uint8Array | string} seedPhrase - The BIP39-compliant seed phrase,
    * either as a string or Uint8Array.
-   * @returns {Promise<Object>} A Promise that resolves to the state.
+   * @returns {Promise<object>} A Promise that resolves to the state.
    */
   createNewVaultAndRestore(password, seedPhrase) {
-    let encodedSeedPhrase = seedPhrase;
-    if (typeof encodedSeedPhrase === 'string') {
-      const indices = seedPhrase
-        .split(' ')
-        .map((word) => wordlist.indexOf(word));
-      encodedSeedPhrase = new Uint8Array(new Uint16Array(indices).buffer);
-    } else if (
-      seedPhrase instanceof Object &&
-      !(seedPhrase instanceof Uint8Array)
-    ) {
-      // when passed from the frontend to the background process a Uint8Array becomes a javascript object
-      encodedSeedPhrase = Uint8Array.from(Object.values(seedPhrase));
-    }
+    const encodedSeedPhrase = this._mnemonicToUint8Array(seedPhrase);
 
     if (typeof password !== 'string') {
       throw new Error('Password must be text.');
     }
 
     if (!bip39.validateMnemonic(encodedSeedPhrase, wordlist)) {
-      return Promise.reject(new Error('Seed phrase is invalid.'));
+      return Promise.reject(
+        new Error('KeyringController - Seed phrase is invalid.'),
+      );
     }
 
     this.password = password;
 
+    this.clearKeyrings();
     return this.persistAllKeyrings(password)
       .then(() => {
         return this.addNewKeyring(KEYRINGS_TYPE_MAP.HD_KEYRING, {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -914,6 +915,52 @@ class KeyringController extends EventEmitter {
...
+    );
+    return recoveredIndices.map((i) => wordlist[i]).join(' ');
+  }
+
+  _stringToUint8Array(mnemonic) {
+    const indices = mnemonic.split(' ').map((word) => wordlist.indexOf(word));
+    return new Uint8Array(new Uint16Array(indices).buffer);
+  }
+
+  _mnemonicToUint8Array(mnemonic) {
```
adonesky1, line 935: Suppose it could be useful to go back and make this a public static method on `eth-hd-keyring` to avoid this duplication. cc @Gudahtt

index.js
```diff
@@ -107,45 +108,45 @@ class KeyringController extends EventEmitter {
...
     }
 
-    const wordlists = Object.values(bip39.wordlists);
-    if (
-      wordlists.every(
-        (wordlist) => !bip39.validateMnemonic(seedPhraseAsBuffer, wordlist),
-      )
-    ) {
-      throw new Error('Seed phrase is invalid.');
+    if (!bip39.validateMnemonic(encodedSeedPhrase, wordlist)) {
```
adonesky1, line 122: We no longer validate against every wordlist.

index.js
```diff
@@ -107,40 +108,30 @@ class KeyringController extends EventEmitter {
...
     }
 
-    const wordlists = Object.values(bip39.wordlists);
-    if (
-      wordlists.every(
-        (wordlist) => !bip39.validateMnemonic(seedPhraseAsBuffer, wordlist),
-      )
-    ) {
-      throw new Error('Seed phrase is invalid.');
+    if (!bip39.validateMnemonic(seedPhraseUint8Array, wordlist)) {
```
adonesky1, line 122: we only validate against the english wordlist now.

index.js
Gudahtt: Good point. It would be ideal to eliminate this duplication.

If we're using this code in both places, maybe it would be better to move this to a third place? e.g. https://github.com/MetaMask/utils or `@metamask/scure-bip39`. Just a thought; `eth-hd-keyring` would suffice for now.

adonesky1: yeah maybe `@metamask/scure-bip39` may make the most sense?

Gudahtt: Or we could drop this validation step. It seems to be redundant; `eth-hd-keyring` is doing it already here: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L284

index.js
```diff
@@ -107,40 +108,30 @@ class KeyringController extends EventEmitter {
...
-    this.clearKeyrings();
-    const firstKeyring = await this.addNewKeyring(
-      KEYRINGS_TYPE_MAP.HD_KEYRING,
-      {
-        mnemonic: seedPhraseAsBuffer,
-        numberOfAccounts: 1,
-      },
-    );
-    const [firstAccount] = await firstKeyring.getAccounts();
+    await this.clearKeyrings();
```
Gudahtt, line 128: Interesting; another `async` call without an `await`. In this case the function wasn't doing anything async, so this should result in no functional change.

index.js
Gudahtt: Makes sense.

For the changelog, is this a breaking change? i.e. did `eth-hd-keyring` ever support non-english mnemonics?

index.js
adonesky1: lol true. In which case `KeyringController` needn't depend directly on `@metamask/scure-bip39` at all!

Gudahtt: Oh! That would be good as well

adonesky1: Removed `bip39` dependency altogether in latest commit: https://github.com/MetaMask/KeyringController/pull/179/commits/dea4b70574e4da2d72882801dbe7db0d1e93f1b9

Gudahtt (approved):
LGTM!

index.js
adonesky1: No it did not. Back when we were using `bitcoinjs/bip39` and our fork [we would validate against](https://github.com/bitcoinjs/bip39/blob/f88d0dd4db96b3ccbba1c8bd6ec87bff54025949/src/index.js#L67) the `default` wordlist which was [set to english by default].(https://github.com/bitcoinjs/bip39/blob/f88d0dd4db96b3ccbba1c8bd6ec87bff54025949/src/_wordlists.js#L57).

Gudahtt: Great, thanks!

PR #193, opened 2023-01-18
Bump `@metamask/eth-hd-keyring` to v6.0.0

adonesky1:
`@metamask/eth-hd-keyring` v5.x is now deprecated. v6.0.0 reverts the serialization format from Uint8Arrays (the format introduced in v5) because they don't serialize to JSON nicely, back to an array of utf8 encoded bytes as it was prior to v5.0.0


Gudahtt (approved):
LGTM!

Issue #195, opened 2023-01-24
Migrate to TypeScript


Gudahtt:
We want to migrate this package to TypeScript to reduce regressions and improve compatibility with our other TypeScript projects.

PR #202, opened 2023-03-01
BREAKING: TS Refactor

gantunesr:
**BREAKING: TS Migration**

This PR introduces the following changes,

- Refactors the KeyringController code and tests to TypeScript.
- Add guards logic to ensure the correct logic flow.
- Updates the package configuration
- Updates error messages

**Description**

_Itemize the changes you have made into the categories below_

- BREAKING:

  - The `addNewKeyring` method now expects an object containing the property `privateKeys` of type `string[]` in the case the keyring to be added is a "Simple Keyring".

- FIXED:

  - Method `getKeyringForAccount`
    - Add `async` keyword for method returning a `Promise`
    - Update conditional statements

- CHANGED:

  - Method `createNewVaultAndRestore`
    - Add guard logic in case `keyring` is undefined.
  - Method `addNewKeyring` 
    - Add guard logic in case `keyring` is undefined.
    - Add guard logic in case the method `generateRandomMnemonic` doesn't exist in the current keyring
  - Method `exportAccount` 
    - Add guard logic in case the method `exportAccount` doesn't exist in the current keyring
  - Method `signTransaction` 
    - Add guard logic in case the method `signTransaction` doesn't exist in the current keyring
  - Method `signMessage` 
    - Add guard logic in case the method `signMessage` doesn't exist in the current keyring
  - Method `signPersonalMessage` 
    - Add guard logic in case the method `signPersonalMessage` doesn't exist in the current keyring
  - Method `getEncryptionPublicKey` 
    - Add guard logic in case the method `getEncryptionPublicKey` doesn't exist in the current keyring
  - Method `decryptMessage` 
    - Add guard logic in case the method `decryptMessage` doesn't exist in the current keyring
  - Method `getAppKeyAddress` 
    - Add guard logic in case the method `getAppKeyAddress` doesn't exist in the current keyring
  - Method `unlockKeyrings` 
    - Add guard logic in case the `encryptionKey` data type is not a string
    - Add guard logic in case the `password` data type is not a string
    - Logic updated to handle errors
  - Method `getKeyringForAccount`
    - Update conditional statements
   - Method `restoreKeyring`
     - Now is a private method 
     - Logic updated to handle errors
   - Method `clearKeyrings`
     - Now is a private method
   - Method `newKeyring`
     - Now is a private method
 
**Checklist**

- [x] Tests are included if applicable
- [x] Any added code is fully documented

**Issue**

Progresses https://github.com/MetaMask/KeyringController/issues/195


src/KeyringController.ts
```diff
@@ -220,16 +252,30 @@ class KeyringController extends EventEmitter {
...
+  async addNewKeyring(
+    type: string,
+    opts?: Record<string, unknown>,
+  ): Promise<ExtendedKeyring | void> {
+    const keyring = await this.#newKeyring(
+      type,
+      type === KeyringType.Simple && opts?.privateKeyArray
+        ? opts?.privateKeyArray
+        : opts,
+    );
```
gantunesr, line 268: Breaking change

src/KeyringController.ts
```diff
@@ -273,26 +325,29 @@ class KeyringController extends EventEmitter {
...
     switch (type) {
-      case KEYRINGS_TYPE_MAP.SIMPLE_KEYRING: {
+      case KeyringType.Simple: {
         const isIncluded = Boolean(
           accounts.find(
-            (key) =>
+            (key: string) =>
               key === newAccountArray[0] ||
-              key === stripHexPrefix(newAccountArray[0]),
+              key === stripHexPrefix(newAccountArray[0] as string),
```
legobeat, line 344:
```suggestion
              // strip hex prefix
              key === parseInt(newAccountArray[0] as string, 16).toString(16),
```

This would remove the need to declare the `HEX_PREFIX` constant, I think?

src/KeyringController.ts
```diff
@@ -220,16 +254,34 @@ class KeyringController extends EventEmitter {
...
-   * @returns {Promise<Keyring>} The new keyring.
+   * @param type - The type of keyring to add.
+   * @param opts - The constructor options for the keyring.
+   * @returns The new keyring.
    */
-  async addNewKeyring(type, opts) {
-    const keyring = await this._newKeyring(type, opts);
+  async addNewKeyring(
+    type: string,
+    opts?: Record<string, unknown>,
```
mcmire, line 263: To save from having to say `?.` should this default to any empty object?

``` suggestion
    opts?: Record<string, unknown> = {},
```

src/KeyringController.ts
```diff
@@ -1,41 +1,60 @@
...
-  //
-  // PUBLIC METHODS
-  //
+  keyringBuilders: { (): Keyring<Json>; type: string }[];
+
+  public store: typeof ObservableStore;
+
+  public memStore: typeof ObservableStore;
+
+  public encryptor: any;
```
mcmire, line 47: Any way we can provide a real type here?

src/KeyringController.ts
```diff
@@ -1,41 +1,60 @@
...
+  public keyrings: Keyring<State>[];
+
+  public cacheEncryptionKey: boolean;
+
+  public unsupportedKeyrings: { type: string; data: Json }[];
 
-  constructor(opts) {
+  public password?: string | undefined;
+
+  constructor(opts: any) {
```
mcmire, line 57: Can we provide a type for `opts` instead of accepting `any`?

src/KeyringController.ts
```diff
@@ -273,26 +325,29 @@ class KeyringController extends EventEmitter {
...
-   * @param {string} type - The key pair type to check for.
-   * @param {Array<string>} newAccountArray - Array of new accounts.
-   * @returns {Promise<Array<string>>} The account, if no duplicate is found.
+   * @param type - The key pair type to check for.
+   * @param newAccountArray - Array of new accounts.
+   * @returns The account, if no duplicate is found.
    */
-  async checkForDuplicate(type, newAccountArray) {
+  async checkForDuplicate(
+    type: string,
```
mcmire, line 333: Do we have a list of valid keyring types somewhere that we can use here instead of accepting anything? If that's not possible right now, no problem, but it might be something we want to think about improving.

src/KeyringController.ts
```diff
@@ -273,26 +325,29 @@ class KeyringController extends EventEmitter {
...
+  ): Promise<string[]> {
     const accounts = await this.getAccounts();
 
     switch (type) {
-      case KEYRINGS_TYPE_MAP.SIMPLE_KEYRING: {
+      case KeyringType.Simple: {
         const isIncluded = Boolean(
           accounts.find(
-            (key) =>
+            (key: string) =>
```
mcmire, line 342: Nit: `getAccounts` returns a string, so typing this shouldn't be necessary.

src/KeyringController.ts
```diff
@@ -442,68 +544,141 @@ class KeyringController extends EventEmitter {
...
-   * @param {string} origin - The origin for the app key.
-   * @returns {string} The app key private key.
+   * @param _address - The Ethereum address for the app key.
+   * @param origin - The origin for the app key.
+   * @returns The app key private key.
    */
-  async exportAppKeyForAddress(_address, origin) {
-    const address = normalizeAddress(_address);
+  async exportAppKeyForAddress(
+    _address: string,
```
mcmire, line 619: Similar as above, what about `rawAddress`?

src/KeyringController.ts
```diff
@@ -442,68 +544,141 @@ class KeyringController extends EventEmitter {
...
+   * Restore Keyring
+   *
+   * Attempts to initialize a new keyring from the provided serialized payload.
+   * On success, updates the memStore keyrings and returns the resulting
+   * keyring instance.
+   *
+   * @param serialized - The serialized keyring.
+   * @returns The deserialized keyring.
+   */
+  async restoreKeyring(serialized: any): Promise<Keyring<State> | undefined> {
```
mcmire, line 646: Is it possible to use `Json` or something here instead of `any`? Or is that not possible because this could theoretically come from anywhere?

src/KeyringController.ts
```diff
@@ -662,48 +861,36 @@ class KeyringController extends EventEmitter {
...
    * Attempts to initialize a new keyring from the provided serialized payload.
    * On success, returns the resulting keyring instance.
    *
-   * @param {object} serialized - The serialized keyring.
-   * @returns {Promise<Keyring|undefined>} The deserialized keyring or undefined if the keyring type is unsupported.
+   * @param serialized - The serialized keyring.
+   * @returns The deserialized keyring or undefined if the keyring type is unsupported.
    */
-  async _restoreKeyring(serialized) {
+  async #restoreKeyring(serialized: any): Promise<Keyring<State> | undefined> {
```
mcmire, line 888: Since `type` and `data` are being destructured out, should we give a type to `serialized`?

src/KeyringController.ts
```diff
@@ -721,40 +908,28 @@ class KeyringController extends EventEmitter {
    * matches the provided `type`,
    * returning it if it exists.
    *
-   * @param {string} type - The type whose class to get.
-   * @returns {Keyring|undefined} The class, if it exists.
+   * @param type - The type whose class to get.
+   * @returns The class, if it exists.
    */
-  getKeyringBuilderForType(type) {
+  #getKeyringBuilderForType(type: string): any {
```
mcmire, line 914: Instead of `any` should this be `{ (): Keyring<Json>; type: string }`?

src/KeyringController.ts
```diff
@@ -785,21 +960,22 @@ class KeyringController extends EventEmitter {
       const accounts = candidate[1].map(normalizeAddress);
       return accounts.includes(hexed);
     });
-    if (winners && winners.length > 0) {
+
+    if (winners?.length && winners?.[0]?.length) {
```
mcmire, line 964: `filter` doesn't return undefined, so do you need the first `?.` here?

``` suggestion
    if (winners.length && winners[0]?.length) {
```

src/KeyringController.ts
```diff
@@ -785,21 +960,22 @@ class KeyringController extends EventEmitter {
...
       return winners[0][0];
     }
 
     // Adding more info to the error
     let errorInfo = '';
     if (!address) {
       errorInfo = 'The address passed in is invalid/empty';
-    } else if (!candidates || !candidates.length) {
+    } else if (!candidates?.length) {
```
mcmire, line 972: It doesn't seem that you used `?.` when referring to `candidates` further up this method, so do you need the `?.` here?

src/KeyringController.ts
```diff
@@ -785,21 +960,22 @@ class KeyringController extends EventEmitter {
...
       errorInfo = 'There are no keyrings';
-    } else if (!winners || !winners.length) {
+    } else if (!winners?.length) {
```
mcmire, line 974: Similar to a previous comment — is `?.` needed?

src/KeyringController.ts
```diff
@@ -826,69 +1004,37 @@ class KeyringController extends EventEmitter {
...
-   * @param {string} type - The type of keyring to add.
-   * @param {object} data - The data to restore a previously serialized keyring.
-   * @returns {Promise<Keyring>} The new keyring.
+   * @param type - The type of keyring to add.
+   * @param data - The data to restore a previously serialized keyring.
+   * @returns The new keyring.
    */
-  async _newKeyring(type, data) {
-    const keyringBuilder = this.getKeyringBuilderForType(type);
+  async #newKeyring(type: string, data: any): Promise<Keyring<State> | void> {
```
mcmire, line 1036: It seems that the argument for `Keyring.deserialize` is of type `State`. Can we use that here instead of `any`?

src/mocks/encryptor.mock.ts
```diff
@@ -16,17 +17,18 @@ const INVALID_PASSWORD_ERROR = 'Incorrect password.';
 const MOCK_HARDCODED_KEY = 'key';
 const MOCK_HEX = '0xabcdef0123456789';
 const MOCK_SALT = 'SALT';
+// eslint-disable-next-line no-restricted-globals
```
mcmire, line 20: It feels like we ought to be able to access Node globals in test-related files. Wondering if there is a change to the ESLint config we can make to allow this. I'll do some research.

src/mocks/encryptor.mock.ts
```diff
@@ -1,10 +1,11 @@
-const sinon = require('sinon');
```
mcmire, line 1: If the mock files are exclusively for tests, what do you think about moving them back to `test`? Otherwise they will get written to `dist/` and included in releases accidentally.

src/mocks/keyring.mock.ts
```diff
@@ -1,23 +1,23 @@
 class KeyringMockWithInit {
```
mcmire, line 1: Is there a way we can ensure this conforms to the keyring interface? Can we say something like this?

``` suggestion
class KeyringMockWithInit implements KeyringClass {
```

src/mocks/keyring.mock.ts
```diff
@@ -1,23 +1,23 @@
...
     return Promise.resolve({});
   }
 
-  deserialize(_) {
+  async deserialize(_) {
     return Promise.resolve();
   }
 }
 
 KeyringMockWithInit.type = 'Keyring Mock With Init';
```
mcmire, line 21: Instead of setting this outside the class, I believe you could set it inside the class like so:

``` typescript
static type = 'Keyring Mock With Init'
```

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "lib": ["ES2020", "dom"],
+    "module": "CommonJS",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noErrorTruncation": true,
+    "noUncheckedIndexedAccess": true,
+    "strict": true,
+    "target": "ES2017",
+    "noImplicitAny": false
```
mcmire, line 14: `noImplicitAny` is already covered by `strict`.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "exclude": ["./dist/**/*", "node_modules"]
```
mcmire, line 16: Good call to add `node_modules` here — we should add this to the module template.

tsconfig.build.json
```diff
@@ -0,0 +1,17 @@
...
+    "inlineSources": true,
+    "noEmit": false,
+    "outDir": "dist",
+    "rootDir": "src",
+    "sourceMap": true
+  },
+  "include": ["./src/**/*.ts"],
+  "exclude": [
+    "./src/**/*.test.ts",
+    "./src/**/*.test-d.ts",
```
mcmire, line 14: Good call to add this one — we should add this to the module template.

mcmire (commented):
Did a first pass on this PR. Spotted some usages of `any` and/or potential opportunities to improve the types. I'll do another pass a bit later.

src/KeyringController.ts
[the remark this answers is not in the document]
gantunesr: We can add a conditional to verify `newAccountArray.length`

src/KeyringController.ts
```diff
@@ -331,12 +386,18 @@ class KeyringController extends EventEmitter {
...
+  async exportAccount(address: Hex): Promise<string> {
     const keyring = await this.getKeyringForAccount(address);
-    return await keyring.exportAccount(normalizeAddress(address));
+    if (!keyring.exportAccount) {
+      throw new Error(
+        `KeyringController - The keyring for address ${address} does not support the method exportAccount.`,
+      );
+    }
+
+    return await keyring.exportAccount(normalizeAddress(address) as Hex);
```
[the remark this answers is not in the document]
gantunesr: Ah, it's expecting a `Hex` only, you're right, augmenting should fix it

src/KeyringController.ts
```diff
@@ -442,68 +544,141 @@ class KeyringController extends EventEmitter {
...
-    return keyring.decryptMessage(address, msgParams.data, opts);
+    if (!keyring.decryptMessage) {
+      throw new Error(
+        `KeyringController - The keyring for address ${address} does not support the method decryptMessage.`,
+      );
+    }
+
+    return keyring.decryptMessage(
+      address,
+      msgParams.data as Eip1024EncryptedData,
```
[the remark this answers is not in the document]
gantunesr: This is the error I'm getting,

```
Argument of type 'string | Eip1024EncryptedData | Record<string, unknown>[]' is not assignable to parameter of type 'Eip1024EncryptedData'.

Type 'string' is not assignable to type 'Eip1024EncryptedData'.
```

A conditional should help to avoid the casting.

src/KeyringController.ts
gantunesr: I have defined this `enum`, we could use that one,

```ts
export enum KeyringType {
  HD = 'HD Key Tree',
  Simple = 'Simple Key Pair',
}
```

src/KeyringController.ts
gantunesr: No need, I guess I just misplaced it

src/KeyringController.ts
gantunesr: No, removing it

tsconfig.json
gantunesr: I have opened a draft PR to update the configuration.

https://github.com/MetaMask/metamask-module-template/pull/181

src/KeyringController.ts
gantunesr: But ultimately, there can be unknown keyrings type outside those

src/KeyringController.ts
```diff
@@ -442,68 +544,141 @@ class KeyringController extends EventEmitter {
...
-   * @param {string} _address - The Ethereum address for the app key.
-   * @param {string} origin - The origin for the app key.
-   * @returns {string} The app key address.
+   * @param _address - The Ethereum address for the app key.
+   * @param origin - The origin for the app key.
+   * @returns The app key address.
    */
-  async getAppKeyAddress(_address, origin) {
-    const address = normalizeAddress(_address);
+  async getAppKeyAddress(_address: string, origin: string): Promise<string> {
```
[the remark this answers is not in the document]
gantunesr: Fixed at [8aa5ba0](https://github.com/MetaMask/KeyringController/pull/202/commits/8aa5ba07f5c2e9a1acba22b2a3d6d47ca646ad95)

src/KeyringController.ts
```diff
@@ -380,30 +441,49 @@ class KeyringController extends EventEmitter {
...
+   * @param opts - Signing options.
+   * @returns The signed transaction object.
    */
-  async signTransaction(ethTx, _fromAddress, opts = {}) {
-    const fromAddress = normalizeAddress(_fromAddress);
-    const keyring = await this.getKeyringForAccount(fromAddress);
-    return await keyring.signTransaction(fromAddress, ethTx, opts);
+  async signTransaction(
+    ethTx: TypedTransaction,
+    _address: string | Hex,
```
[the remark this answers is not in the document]
gantunesr: Fixed at [8aa5ba0](https://github.com/MetaMask/KeyringController/pull/202/commits/8aa5ba07f5c2e9a1acba22b2a3d6d47ca646ad95)

src/KeyringController.ts
gantunesr: Fixed at [8aa5ba0](https://github.com/MetaMask/KeyringController/pull/202/commits/8aa5ba07f5c2e9a1acba22b2a3d6d47ca646ad95)

src/KeyringController.ts
gantunesr: Fixed at [f609772](https://github.com/MetaMask/KeyringController/pull/202/commits/f609772065886c279c181bb532dfe5cdb4c94751)

src/KeyringController.ts
gantunesr: Removed at [f609772](https://github.com/MetaMask/KeyringController/pull/202/commits/f609772065886c279c181bb532dfe5cdb4c94751)

declarations.d.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/unambiguous
```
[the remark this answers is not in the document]
gantunesr: Sounds like a good approach. [1d38cfc](https://github.com/MetaMask/KeyringController/pull/202/commits/1d38cfc72fecec992244815ed2506065d65c8427)

src/KeyringController.ts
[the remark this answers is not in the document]
gantunesr: Improved at [a96cbca](https://github.com/MetaMask/KeyringController/pull/202/commits/a96cbca2a56dae5add872959dc8432488c4a056b).

src/KeyringController.ts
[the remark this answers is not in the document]
gantunesr: What do you about creating a method to check if the data type is `Eip1024EncryptedData` in `@metamask/utils`? Something like,

```ts
isEip1024EncryptedData(data: unknown): boolean
```

src/KeyringController.ts
Gudahtt: Good question. Using `Json` here would make sense to me.

It looks like this comes from the decryptor, which types this as `unknown`. But it should be guaranteed to be JSON in practice. See here for extension: https://github.com/MetaMask/browser-passworder/blob/main/src/index.ts#L176 and mobile: https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Encryptor.js#L75

src/KeyringController.ts
gantunesr: Not sure what's the best approach here since `@metamask/browser-passworder` is providing a series of methods. I can try to define a type like this one,

```ts
export type Encryptor = {
  encrypt<R>(
    password: string,
    dataObj: R,
    key?: CryptoKey,
    salt?: string,
  ): Promise<string>;

  decrypt(password: string, text: string, key?: CryptoKey): Promise<unknown>;

  encryptWithDetail<R>(
    password: string,
    dataObj: R,
    salt?: string,
  ): Promise<DetailedEncryptionResult>;

  decryptWithDetail(
    password: string,
    text: string,
  ): Promise<DetailedDecryptResult>;

  encryptWithKey<R>(key: CryptoKey, dataObj: R): Promise<EncryptionResult>;

  decryptWithKey<R>(key: CryptoKey, payload: EncryptionResult): Promise<R>;

  importKey(keyString: string): Promise<CryptoKey>;
};
```

wdyt?

Gudahtt: `encrypt` and `decrypt` should be the only mandatory functions; those are the ones relied upon by mobile.

The other methods are used only when the `cacheEncyptionKey` option is enabled. If that is enabled they are required, but otherwise we should accept the simpler encryptor to preserve type compatibility with mobile.

Gudahtt: Hmm, maybe we could make the class generic? So we can ensure the extra methods are required for extension but not for mobile.

src/KeyringController.ts
gantunesr: I have defined the new data type `SerializedKeyring` in [ce7d126](https://github.com/MetaMask/KeyringController/pull/202/commits/ce7d126375790bd07789a82f6c9609f0a2eb9813). Let me know what you think

src/KeyringController.ts
gantunesr: Improved in [ce7d126](https://github.com/MetaMask/KeyringController/pull/202/commits/ce7d126375790bd07789a82f6c9609f0a2eb9813)

src/KeyringController.ts
gantunesr: Improved in [ce7d126](https://github.com/MetaMask/KeyringController/pull/202/commits/ce7d126375790bd07789a82f6c9609f0a2eb9813).

src/mocks/encryptor.mock.ts
gantunesr: Agreed [352b11c](https://github.com/MetaMask/KeyringController/pull/202/commits/352b11cab967de2377dd6b7ac570218660126997)

src/mocks/keyring.mock.ts
gantunesr: [476124b](https://github.com/MetaMask/KeyringController/pull/202/commits/476124b7ba759f5880f79cdd5e565419473b4600)

src/KeyringController.ts
gantunesr: [d0f6e63](https://github.com/MetaMask/KeyringController/pull/202/commits/d0f6e635771905e76c0e6f6d47eb874e8f28d16a)

package.json
```diff
@@ -18,31 +18,42 @@
...
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
     "test": "jest"
   },
   "dependencies": {
+    "@ethereumjs/tx": "^4.1.1",
     "@metamask/browser-passworder": "^4.0.2",
     "@metamask/eth-hd-keyring": "^6.0.0",
     "@metamask/eth-sig-util": "5.0.2",
     "@metamask/eth-simple-keyring": "^5.0.0",
+    "@metamask/utils": "5.0.0",
```
leotm, line 42:
```suggestion
    "@metamask/utils": "^5.0.0",
```

or pin all deps

package.json
```diff
@@ -51,10 +62,14 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.2.1",
     "ethereumjs-wallet": "^1.0.1",
-    "jest": "^27.0.6",
+    "jest": "^29.0.0",
     "prettier": "^2.8.1",
     "prettier-plugin-packagejson": "^2.3.0",
-    "sinon": "^11.1.1"
+    "sinon": "15.0.1",
```
leotm, line 68:
```suggestion
    "sinon": "^15.0.1",
```

or pin all deps

package.json
```diff
@@ -51,10 +62,14 @@
...
+    "ts-jest": "^29.0.3",
+    "ts-node": "^10.7.0",
+    "typedoc": "^0.23.15",
+    "typescript": "~4.8.4"
```
leotm, line 72: nb: `4.9.3-5` stable vers also out and `5.0.0` stable due [today](https://github.com/microsoft/TypeScript/issues/51362)
but if problematic can always upgrade to these later

package.json
gantunesr: [b76b3db](https://github.com/MetaMask/KeyringController/pull/202/commits/b76b3db48908afa046642af309b3665708466736)

package.json
gantunesr: [b76b3db](https://github.com/MetaMask/KeyringController/pull/202/commits/b76b3db48908afa046642af309b3665708466736)

types/@metamask/eth-sig-util.d.ts
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line import/unambiguous
+declare module '@metamask/eth-sig-util' {
+  import { Hex } from '@metamask/utils';
+
+  function normalize(address: string | Hex): Hex;
+  export { normalize };
```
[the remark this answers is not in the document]
leotm: fair enough only `const { normalize: normalizeAddress } = require('@metamask/eth-sig-util')` atm

don't have the context to know if will be used, but the export can be added when needed ^

leotm (commented):
i'll bump us to TypeScript v5 since its out

including @typescript-eslint/parser 5.53.0 -> 5.55.0 pre-requisite
(for official ESLint TS5 support i.e. no warnings thrown)

and reference related stuff in commits

leotm:
nb: gotcha - `git clean -fdx && yarn && yarn lint` required to verify ESLint-TypeScript compat

since ESLint caches the TypeScript compat warning, demo'd below

Screenshot 2023-03-17 at 13 57 37

fixed since in https://github.com/MetaMask/KeyringController/pull/202/commits/850cb00785bd58f4bad757ca2552fcb20372e5bc

leotm (dismissed):

.yarnrc.yml
```diff
@@ -10,6 +10,6 @@ nodeLinker: node-modules
 
 plugins:
   - path: .yarn/plugins/@yarnpkg/plugin-allow-scripts.cjs
-    spec: "https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js"
+    spec: 'https://raw.githubusercontent.com/LavaMoat/LavaMoat/main/packages/yarn-plugin-allow-scripts/bundles/@yarnpkg/plugin-allow-scripts.js'
```
mcmire, line 13: A change to this file would seem to indicate that we haven't configured Prettier to ignore this file. I just checked and we do seem to be making an exception for this file (meaning that it should not be getting linted). Would you mind reverting this change? It would get blown away anyway if we added a new Yarn plugin or otherwised updated the Yarn config in any way.

src/KeyringController.test.ts
```diff
@@ -1,17 +1,19 @@
...
-const sinon = require('sinon');
-
-const { KeyringController, keyringBuilderFactory } = require('..');
-const {
+import HdKeyring from '@metamask/eth-hd-keyring';
+import { normalize as normalizeAddress } from '@metamask/eth-sig-util';
+import type { Hex, Json, Keyring, KeyringClass } from '@metamask/utils';
+import { strict as assert } from 'assert';
+import Wallet from 'ethereumjs-wallet';
+import { restore, spy, stub, assert as sinonAssert } from 'sinon';
```
mcmire, line 6: It might be nice to import these as a group here for clarity. We could probably guess that `spy` and `stub` are coming from Sinon, but `restore` is rather ambiguous, and `assert` is a built-in module (you've renamed it of course, but maybe that's a sign). What do you think about:

``` typescript
import * as sinon from 'sinon';
```

src/KeyringController.test.ts
```diff
@@ -65,56 +72,59 @@ describe('KeyringController', function () {
...
+    it('should not load keyrings when incorrect password', async () => {
       await keyringController.createNewVaultAndKeychain(PASSWORD);
       await keyringController.persistAllKeyrings();
       expect(keyringController.keyrings).toHaveLength(1);
 
       await keyringController.setLocked();
 
       await expect(
-        keyringController.submitPassword(`${PASSWORD}a`),
+        keyringController.submitPassword(`${PASSWORD as string}a`),
```
mcmire, line 94: Any way to avoid this typecast? Seems a bit odd.

src/KeyringController.test.ts
```diff
@@ -65,56 +72,59 @@ describe('KeyringController', function () {
...
+  describe('persistAllKeyrings', () => {
+    it('should persist keyrings in _unsupportedKeyrings array', async () => {
+      const unsupportedKeyring = { type: 'DUMMY_KEYRING', data: {} };
+      keyringController.unsupportedKeyrings = [unsupportedKeyring];
       await keyringController.persistAllKeyrings();
 
       const { vault } = keyringController.store.getState();
-      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault);
-      expect(keyrings.indexOf(unsupportedKeyring) > -1).toBe(true);
+      const keyrings = (await mockEncryptor.decrypt(PASSWORD, vault)) as {
```
mcmire, line 118: Any way to avoid this typecast?

src/KeyringController.test.ts
```diff
@@ -266,14 +278,17 @@ describe('KeyringController', function () {
...
     });
 
-    it('throws error if argument password is not a string', async function () {
-      await expect(() =>
-        keyringController.createNewVaultAndRestore(12, walletTwoSeedWords),
-      ).rejects.toThrow('Password must be text.');
+    it('throws error if argument password is not a string', async () => {
+      await expect(async () =>
+        keyringController.createNewVaultAndRestore(
+          12 as any,
```
mcmire, line 284: At some point it would be nice to enable a lint rule that prevents the use of `any`. To anticipate this, what do you think about using `@ts-expect-error` to tell TypeScript it's okay that this type doesn't match?

src/KeyringController.test.ts
```diff
@@ -360,62 +377,76 @@ describe('KeyringController', function () {
...
+    it('should add HD Key Tree with mnemonic passed as an argument', async () => {
       const previousAllAccounts = await keyringController.getAccounts();
       expect(previousAllAccounts).toHaveLength(1);
-      const keyring = await keyringController.addNewKeyring('HD Key Tree', {
+      const keyring = await keyringController.addNewKeyring(KeyringType.HD, {
         numberOfAccounts: 2,
         mnemonic: walletTwoSeedWords,
       });
-      const keyringAccounts = await keyring.getAccounts();
+      const keyringAccounts = await keyring?.getAccounts();
```
mcmire, line 397: Hmm. If we can't guarantee `addNewKeyring` returns a keyring object then what is the point of this test?

src/KeyringController.ts
```diff
@@ -1,45 +1,61 @@
...
- * @param {string} address - The address that might be hex prefixed.
- * @returns {string} The address without a hex prefix.
- */
-function stripHexPrefix(address) {
-  if (address.startsWith('0x')) {
-    return address.slice(2);
-  }
-  return address;
-}
+  public store: typeof ObservableStore;
```
mcmire, line 36: Isn't ObservableStore a class? It seems like we don't need to use `typeof` here.

src/KeyringController.ts
```diff
@@ -1,45 +1,61 @@
...
+  public encryptor: any;
 
-  constructor(opts) {
+  public keyrings: Keyring<Json>[];
+
+  public cacheEncryptionKey: boolean;
+
+  public unsupportedKeyrings: SerializedKeyring[];
+
+  public password?: string | undefined;
```
mcmire, line 48: Nit: I think the `| undefined` happens via `?`:

``` suggestion
  public password?: string;
```

src/KeyringController.ts
```diff
@@ -273,27 +327,28 @@ class KeyringController extends EventEmitter {
...
 
     switch (type) {
-      case KEYRINGS_TYPE_MAP.SIMPLE_KEYRING: {
+      case KeyringType.Simple: {
         const isIncluded = Boolean(
           accounts.find(
             (key) =>
               key === newAccountArray[0] ||
-              key === stripHexPrefix(newAccountArray[0]),
+              key === stripHexPrefix(newAccountArray[0] as Hex),
```
mcmire, line 346: How do we know that `newAccountArray[0]` satisfies `Hex`? Instead of typecasting this, is there a function we can define/use that takes a string and returns a Hex?

src/KeyringController.ts
```diff
@@ -310,17 +365,19 @@ class KeyringController extends EventEmitter {
...
+   * @param selectedKeyring - The currently selected keyring.
+   * @returns A Promise that resolves to the state.
    */
-  async addNewAccount(selectedKeyring) {
+  async addNewAccount(
+    selectedKeyring: Keyring<Json>,
+  ): Promise<KeyringControllerState> {
     const accounts = await selectedKeyring.addAccounts(1);
-    accounts.forEach((hexAccount) => {
+    accounts.forEach((hexAccount: string) => {
```
mcmire, line 375: Why is the type necessary here? TypeScript should be able to infer this.

src/KeyringController.ts
```diff
@@ -412,28 +486,46 @@ class KeyringController extends EventEmitter {
...
     const address = normalizeAddress(msgParams.from);
     const keyring = await this.getKeyringForAccount(address);
-    return await keyring.signPersonalMessage(address, msgParams.data, opts);
+    if (!keyring.signPersonalMessage) {
+      throw new Error(KeyringControllerError.UnsupportedSignPersonalMessage);
+    }
+
+    return await keyring.signPersonalMessage(
+      address,
+      msgParams.data as Hex,
```
mcmire, line 505: Similar to other comments, is there any way we can avoid this typecast?

src/KeyringController.ts
```diff
@@ -412,28 +486,46 @@ class KeyringController extends EventEmitter {
...
-   * @param {object} address - The address to get the encryption public key for.
-   * @param {object} opts - Additional encryption options.
-   * @returns {Promise<Buffer>} The public key.
+   * @param address - The address to get the encryption public key for.
+   * @param opts - Additional encryption options.
+   * @returns The public key.
    */
-  async getEncryptionPublicKey(address, opts = {}) {
+  async getEncryptionPublicKey(
+    address: string | Hex,
```
mcmire, line 520: Just an FYI, this will probably get collapsed to `string` if `Hex` is a subtype of `string`.

src/KeyringController.ts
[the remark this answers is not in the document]
mcmire: If we think it will be applicable to other libraries then it would make sense to put in `utils`, but it seems fairly specific, so maybe placing it in this library would be a good first spot.

In terms of how such a function should be typed, I would probably recommend this instead:

``` typescript
isEip1024EncryptedData(data: unknown): data is Eip1024EncryptedData
```

src/KeyringController.ts
```diff
@@ -808,10 +964,12 @@ class KeyringController extends EventEmitter {
    *
    * Is used for adding the current keyrings to the state object.
    *
-   * @param {Keyring} keyring - The keyring to display.
-   * @returns {Promise<object>} A keyring display object, with type and accounts properties.
+   * @param keyring - The keyring to display.
+   * @returns A keyring display object, with type and accounts properties.
    */
-  async displayForKeyring(keyring) {
+  async #displayForKeyring(
```
mcmire, line 970: Since this method doesn't access `this`, should it be pulled out into a function? That would avoid the need to use `// eslint-disable-next-line @typescript-eslint/unbound-method` when it's used above.

src/KeyringController.ts
```diff
@@ -785,21 +842,120 @@ class KeyringController extends EventEmitter {
...
+   *
+   * Searches the current `keyringBuilders` array
+   * for a Keyring builder whose unique `type` property
+   * matches the provided `type`,
+   * returning it if it exists.
+   *
+   * @param type - The type whose class to get.
+   * @returns The class, if it exists.
+   */
+  #getKeyringBuilderForType(type: string): any {
```
mcmire, line 956: If we truly don't know what this returns, should this be `unknown` rather than `any`?

mcmire (commented):
Another pass, mainly focused on standards around TypeScript. I didn't have enough time to directly compare the JS and TS implementations, but I will try to get to that tomorrow.

.yarnrc.yml
gantunesr: I have added a `prettierignore` file and reverted the changes

[b618965](https://github.com/MetaMask/KeyringController/pull/202/commits/b6189656984c6dcdaf7e511204038fb92beccb5b)

src/KeyringController.test.ts
gantunesr: Thats a good point [a12ce45](https://github.com/MetaMask/KeyringController/pull/202/commits/a12ce456af2831238844dcc5cab8cda903378b2c)

src/KeyringController.test.ts
gantunesr: [91fd437](https://github.com/MetaMask/KeyringController/pull/202/commits/91fd4378dfc92de33b9b8203b39845157eac9244)

src/KeyringController.test.ts
gantunesr: [5b17f35](https://github.com/MetaMask/KeyringController/pull/202/commits/5b17f35678083e8e0cc27688f698e3fe5a9a4bdc)

src/KeyringController.ts
gantunesr: [9c59391](https://github.com/MetaMask/KeyringController/pull/202/commits/9c59391665f98e2b2e75f881204179b17b2248d1)

src/KeyringController.ts
gantunesr: [e36c134](https://github.com/MetaMask/KeyringController/pull/202/commits/e36c1348aadc882625deda681912d6b45786060e)

src/KeyringController.ts
gantunesr: [6eff4a1](https://github.com/MetaMask/KeyringController/pull/202/commits/6eff4a1b4d09b68cc4ec3fc3b4d0ee378f153f90)

src/KeyringController.ts
gantunesr: [e36c134](https://github.com/MetaMask/KeyringController/pull/202/commits/e36c1348aadc882625deda681912d6b45786060e)

src/KeyringController.ts
gantunesr: [e36c134](https://github.com/MetaMask/KeyringController/pull/202/commits/e36c1348aadc882625deda681912d6b45786060e)

src/KeyringController.ts
gantunesr: [7044d64](https://github.com/MetaMask/KeyringController/pull/202/commits/7044d640adbd718a1c8540f8ec3faa1940db8edc)

mcmire (commented):
Did a second pass comparing the JS and TS implementations. Looks like there's still one comment from me and @Gudahtt that is outstanding, but other than that this all seems to make sense to me, so nice job. I just found a few more things other than that, mostly minor.

One thing I noticed is that some of the methods that were previously public (but placed in a private section) are now actually private. Is it worth noting these in the PR description?

src/KeyringController.ts
```diff
@@ -0,0 +1,1066 @@
...
+
+    if (!keyringBuilder) {
+      throw new Error(
+        `${KeyringControllerError.NoKeyringBuilder}. Keyring type: ${type}`,
+      );
+    }
+
+    const keyring = keyringBuilder();
+
+    await keyring.deserialize(data as Json);
```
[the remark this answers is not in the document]
gantunesr: I just noticed there's a `isValidJson` method in `@metamask/utils`. That should do the trick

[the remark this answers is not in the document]
gantunesr: I just noticed there's a isValidJson method in @metamask/utils. That should do the trick

types/@metamask/eth-sig-util.d.ts
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line import/unambiguous
+declare module '@metamask/eth-sig-util' {
+  import { Hex } from '@metamask/utils';
+
+  function normalize(address: string | Hex): Hex;
```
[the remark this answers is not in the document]
gantunesr: This is no longer necessary after all the previous changes [ee755fd](https://github.com/MetaMask/KeyringController/pull/202/commits/ee755fd7534dfcd5417407b1b7bf30d31b8ce9d1)

src/KeyringController.ts
```diff
@@ -0,0 +1,1066 @@
...
+   * @param serialized.data - Keyring data.
+   * @returns The deserialized keyring or undefined if the keyring type is unsupported.
+   */
+  async #restoreKeyring(
+    serialized: SerializedKeyring,
+  ): Promise<Keyring<Json> | undefined> {
+    const { type, data } = serialized;
+
+    let keyring: Keyring<Json> | undefined;
+    try {
```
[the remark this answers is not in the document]
gantunesr: I think the tests,

- `should return undefined if keyring type is not supported.`
- `add serialized keyring to unsupportedKeyrings array if keyring type is not known`

should already take those cases into consideration.

I'll update the PR description to reflect this.

src/KeyringController.ts
```diff
@@ -0,0 +1,1066 @@
...
+  async removeAccount(address: Hex): Promise<KeyringControllerState> {
+    const keyring = await this.getKeyringForAccount(address);
+
+    // Not all the keyrings support this, so we have to check
+    if (typeof keyring.removeAccount === 'function') {
+      keyring.removeAccount(address);
+      this.emit('removedAccount', address);
+    } else {
+      throw new Error(KeyringControllerError.UnsupportedRemoveAccount);
+    }
```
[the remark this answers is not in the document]
gantunesr: [c87885c](https://github.com/MetaMask/KeyringController/pull/202/commits/c87885c11a1bd0e716d1dccde1cf0afe3c5e3498)

src/KeyringController.ts
```diff
@@ -0,0 +1,1066 @@
...
+   * @param origin - The origin for the app key.
+   * @returns The app key private key.
+   */
+  async exportAppKeyForAddress(
+    rawAddress: string,
+    origin: string,
+  ): Promise<string> {
+    const address = normalizeToHex(rawAddress);
+    const keyring = await this.getKeyringForAccount(address);
+    // The "in" operator is typically restricted because it also checks inherited properties,
```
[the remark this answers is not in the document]
gantunesr: [1c76cf0](https://github.com/MetaMask/KeyringController/pull/202/commits/1c76cf097f7f149317c38b6029e5f1c2a4e1d9d0)

src/KeyringController.ts
```diff
@@ -0,0 +1,1066 @@
...
+  async #restoreKeyring(
+    serialized: SerializedKeyring,
+  ): Promise<Keyring<Json> | undefined> {
+    const { type, data } = serialized;
+
+    let keyring: Keyring<Json> | undefined;
+    try {
+      keyring = await this.#newKeyring(type, data);
+    } catch {
+      // Error
```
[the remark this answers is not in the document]
gantunesr: [a344342](https://github.com/MetaMask/KeyringController/pull/202/commits/a34434283fb67a978690bc320f698a20db2b8e77)

gantunesr:
Hey @mcmire, 

Thanks for the review, I still wasn't able to dedicate time to find the best solution for the unresolved comments,

- `encryptor` type (https://github.com/MetaMask/KeyringController/pull/202#discussion_r1130086803)
- The `ObservableStore` comment, this one particularly is throwing this error   `Cannot use namespace 'ObservableStore' as a type.`

mcmire:
@gantunesr Okay no problem! Thanks for trying. Maybe we can address those in the future.

src/KeyringController.ts
mcmire: I'll resolve this conversation since there must be something here I don't quite understand, so this is fine.

src/KeyringController.ts
gantunesr: I have created a ticket to reflect the work to design and develop the data type for `encryptor` https://github.com/MetaMask/KeyringController/issues/206

src/KeyringController.test.ts
```diff
@@ -527,48 +556,54 @@ describe('KeyringController', function () {
...
     });
   });
 
-  describe('verifyPassword', function () {
+  describe('verifyPassword', () => {
     beforeEach(() => {
       keyringController = new KeyringController({
+        keyringBuilders: [
+          keyringBuilderFactory(
+            KeyringMockWithInit as unknown as KeyringClass<any>,
```
mcmire, line 597: Why typecast this? It seems like a mock keyring that we create should conform to the keyring class interface.

mcmire (dismissed):
Noticed some more typecasts in the tests. We should also try to remove those in future PRs. I was curious about one of them, but it isn't blocking. LGTM.

src/KeyringController.test.ts
gantunesr: It should. Done in [977facf](https://github.com/MetaMask/KeyringController/pull/202/commits/977facf11c37f5418c5303954037d7ecf125b734).

I also removed the rest of typecasting in this module.

mcmire (approved):

PR #203, opened 2023-03-03
fix keeping the wallet unlocked on sw restart after onboarding

pedronfigueiredo:
## Explanation

We want to keep the wallet unlocked when a service worker restarts.

In `persistAllKeyrings`, when we're building for mv3 and want to cache the `encryptionKey` but a password was entered and a new `encryptionKey` has been generated, we want to save `encryptionSalt` alongside the new `encryptionKey` in the  keyring controller `memStore`.

This is so it can be [saved in storage](https://github.com/MetaMask/metamask-extension/blob/987daee85483c38ec9592a05fb8efe1f0e71eda7/app/scripts/metamask-controller.js#L4257) and then [retrieved](https://github.com/MetaMask/metamask-extension/blob/987daee85483c38ec9592a05fb8efe1f0e71eda7/app/scripts/metamask-controller.js#L2595) on the metamask controller for the extension.

Note that this PR only fixes the first time the service worker is restarted after a fresh wallet is loaded and the onboarding flow is complete. Unlocked wallet persistence for subsequent service worker restarts has been fixed on a [separate PR](https://github.com/MetaMask/metamask-extension/pull/17950) on the extension repository.

For context, this was [previously addressed](https://github.com/MetaMask/metamask-extension/pull/15558), but the wallet was still locked after a long enough waiting period after the service worker restarted. [Another PR](https://github.com/MetaMask/metamask-extension/pull/16559) has since been merged, making the wallet lock more immediately apparent.

## Manual Testing Steps

1. Build the extension wallet from develop

`git checkout develop && git pull && yarn && yarn:start:mv3`

2. Load the wallet in the browser and complete the onboarding flow

3. Terminate the service worker using `chrome://inspect/#service-workers`. 

4. The wallet should lock after a couple of seconds.

5. Add the changes on `index.js` on this PR to the corresponding file on the `node_modules` folder of the Metamask Extension.

6. Build the wallet again.

7. Load the wallet in the browser and complete the onboarding flow

8. Terminate the service worker using `chrome://inspect/#service-workers`. 

9. The wallet should remain unlocked.

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -10,7 +10,6 @@ const {
   PASSWORD,
   MOCK_HARDCODED_KEY,
   MOCK_HEX,
-  MOCK_SALT,
 } = require('./lib/mock-encryptor');
 const { KeyringMockWithInit } = require('./lib/mock-keyring');
 
@@ -154,32 +153,8 @@ describe('KeyringController', function () {
 
       it('should save an up to date encryption salt to the `memStore` when `password` is set through `createNewVaultAndKeychain`', async function () {
         keyringController.cacheEncryptionKey = true;
-        const vaultEncryptionKey = '🔑';
-        const vaultEncryptionSalt = MOCK_HEX;
-        const vault = JSON.stringify({ salt: vaultEncryptionSalt });
-        keyringController.store.updateState({ vault });
 
         await keyringController.createNewVaultAndKeychain(PASSWORD);
-
-        expect(keyringController.memStore.getState().encryptionKey).toBe(
-          MOCK_HARDCODED_KEY,
-        );
-        expect(keyringController.memStore.getState().encryptionSalt).toBe(
-          MOCK_HEX,
-        );
-
-        await keyringController.unlockKeyrings(
-          undefined,
-          vaultEncryptionKey,
-          vaultEncryptionSalt,
-        );
-
-        expect(keyringController.memStore.getState().encryptionKey).toBe(
-          vaultEncryptionKey,
-        );
-        expect(keyringController.memStore.getState().encryptionSalt).toBe(
-          vaultEncryptionSalt,
-        );
 
         const response = await keyringController.persistAllKeyrings();
 
@@ -194,32 +169,8 @@ describe('KeyringController', function () {
 
       it('should save an up to date encryption salt to the `memStore` when `password` is set through `submitPassword`', async function () {
         keyringController.cacheEncryptionKey = true;
-        const vaultEncryptionKey = '🔑';
-        const vaultEncryptionSalt = MOCK_HEX;
-        const vault = JSON.stringify({ salt: vaultEncryptionSalt });
-        keyringController.store.updateState({ vault });
 
         await keyringController.submitPassword(PASSWORD);
-
-        expect(keyringController.memStore.getState().encryptionKey).toBe(
-          MOCK_ENCRYPTION_KEY,
-        );
-        expect(keyringController.memStore.getState().encryptionSalt).toBe(
-          MOCK_SALT,
-        );
-
-        await keyringController.unlockKeyrings(
-          undefined,
-          vaultEncryptionKey,
-          vaultEncryptionSalt,
-        );
-
-        expect(keyringController.memStore.getState().encryptionKey).toBe(
-          vaultEncryptionKey,
-        );
-        expect(keyringController.memStore.getState().encryptionSalt).toBe(
-          vaultEncryptionSalt,
-        );
 
         const response = await keyringController.persistAllKeyrings();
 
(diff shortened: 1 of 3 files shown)


test/index.js
```diff
@@ -105,10 +108,48 @@ describe('KeyringController', function () {
       await keyringController.persistAllKeyrings();
 
       const { vault } = keyringController.store.getState();
-      const keyrings = await mockEncryptor.decrypt(password, vault);
+      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault);
       expect(keyrings.indexOf(unsupportedKeyring) > -1).toBe(true);
       expect(keyrings).toHaveLength(2);
     });
+
+    describe('on a manifest v3 build', function () {
```
Gudahtt, line 116: Nit: Maybe better to use the same terminology this project uses:

```suggestion
    describe(`when 'cacheEncryptionKey' is enabled`, function () {
```

test/index.js
```diff
@@ -105,10 +108,48 @@ describe('KeyringController', function () {
...
+        );
+
+        expect(keyringController.memStore.getState().encryptionKey).toBe(
+          vaultEncryptionKey,
+        );
+        expect(keyringController.memStore.getState().encryptionSalt).toBe(
+          vaultEncryptionSalt,
+        );
+
+        const response = await keyringController.persistAllKeyrings();
```
Gudahtt, line 142: It looks like this is testing the case in `persistAllKeyrings` where `password` is unset but `encryptionKey` is set. While it's great to have coverage of this scenario, the one that was causing the extension bug is the case where `password` **is** set.

We could get to that state in one of two ways: either call `createNewVaultAndKeychain` or `submitPassword` during the "prepare" step of the test. After either of those calls, `persistAllKeyrings` should go down the path we've changed in this PR.

Gudahtt: It'd be nice to add a test case under `createNewVaultAndKeychain` as well, to confirm that the encryption key and salt are both saved. Likewise for `createNewVaultAndRestore`

Gudahtt (commented):
Looks great! Great to have these additional tests, but I think the code being changed here isn't yet adequately covered. I can approve once it's covered by at least one test

test/index.js
```diff
@@ -105,10 +109,128 @@ describe('KeyringController', function () {
...
+
+        expect(response).toBe(true);
+        expect(keyringController.memStore.getState().encryptionKey).toBe(
+          MOCK_HARDCODED_KEY,
```
Gudahtt, line 147: It's surprising to me that the key would change as a result of `persistAllKeyrings` being called. Having trouble following what's going on here.

Gudahtt: The password is set in the `beforeEach`. The better fix for this would be to move the controller initialization to a setup function, and call it explicitly in each test so that we can customize it when we need to.

But as a simpler workaround, you could unset the password at the beginning of the test, as we're already doing in the "triggers an error when trying to persist without password or encryption key" test

pedronfigueiredo: That makes sense, Mark. Thanks a lot for catching that

test/index.js
```diff
@@ -105,10 +109,129 @@ describe('KeyringController', function () {
...
+          vaultEncryptionKey,
+          vaultEncryptionSalt,
+        );
+
+        expect(keyringController.memStore.getState().encryptionKey).toBe(
+          vaultEncryptionKey,
+        );
+        expect(keyringController.memStore.getState().encryptionSalt).toBe(
+          vaultEncryptionSalt,
+        );
```
Gudahtt, line 182: Nit: The extra `unlockKeyrings` step in these tests makes it hard to tell what they're trying to assert. These two tests would be simpler without that step, and without the associated vault state.

It's a little confusing as well because with a real encryptor, the `unlockKeyrings` step would fail to decrypt the vault because it's using the wrong key. This test shows a key and a password that don't match being used to decrypt the same vault, which should be impossible with a working decryptor.

```suggestion
        await keyringController.createNewVaultAndKeychain(PASSWORD);
        expect(keyringController.memStore.getState().encryptionKey).toBe(
          MOCK_HARDCODED_KEY,
        );
        expect(keyringController.memStore.getState().encryptionSalt).toBe(
          MOCK_HEX,
        );
```

The same simplification can be made to the test immediately following this one as well

Gudahtt (dismissed):
LGTM! Left a comment on two tests that can still be improved

brad-decker (approved):

PR #217, opened 2023-04-18
[FIX] Property name `isUnlocked`

gantunesr:
Change property name from `vaultUnlock` to `isUnlocked`. This was a change from the TS refactor that got merged unintentionally.


Gudahtt (approved):
LGTM!

PR #219, opened 2023-04-19
Bump `ts-jest`

gantunesr:
There is a false positive in the current CI caused by using `TypeScript v.5.0.2` and `ts-jest  v29.0.3`. This development updates it to the required version and ignores the error in the meanwhile.

Screenshot 2023-04-19 at 13 45 21

The PR https://github.com/MetaMask/utils/pull/99 should solve the root cause of the TS error.

### DIFF
--- a/src/KeyringController.ts
+++ b/src/KeyringController.ts
@@ -1036,7 +1036,9 @@ class KeyringController extends EventEmitter {
 
     await keyring.deserialize(data);
 
+    // @ts-expect-error The method 'init' is not part of the current Keyring type
     if (keyring.init) {
+      // @ts-expect-error The method 'init' is not part of the current Keyring type
       await keyring.init();
     }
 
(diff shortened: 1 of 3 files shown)


danroc (approved):

src/KeyringController.ts
```diff
@@ -1036,7 +1036,9 @@ class KeyringController extends EventEmitter {
 
     await keyring.deserialize(data);
 
+    // @ts-expect-error The method 'init' is not part of the current Keyring type
     if (keyring.init) {
+      // @ts-expect-error The method 'init' is not part of the current Keyring type
```
Gudahtt, line 1041: Nit: We can check for this function in a way that TypeScript understands, rather than suppress the error:

```suggestion
    if (hasProperty(keyring, 'init') && typeof keyring.init === 'function') {
```

The `hasProperty` function used here would have to be imported, but it's in `@metamask/utils` which is already a dependency.

Gudahtt (approved):
LGTM! one suggestion though

src/KeyringController.ts
gantunesr: I'll open another PR to include your suggestion @Gudahtt

PR #221, opened 2023-04-19
Improve `build-lint-test` workflow

gantunesr:

PR #228, opened 2023-05-12
[FIX] Inconsistent data types and validations

gantunesr:
## Description

The TS refactor and workflow update had issues that prevented the module to be correctly published and consumed by the clients, this PR aims to solve them.

The noticeable issues were,

1. The method `fullUpdate` was marked as private when it should be public.
2. The method `normalizeToHex` is returning a `string` to methods that require a `Hex`.
3. The data type validation for the parameter `data` in the `#newKeyring` was removed because the clients are not sending the correct input. The validation will be re-introduced after the proper updates in the clients.
4. The `tsconfig.build.json` file was update in accordance to the module template.
5. the `.d.ts` files were misplaced in the wrong directory

## Changes

  
- **CHANGED**:
  - The methods that consume the method `normalizeToHex` have to cast the return type as `Hex`.
  - The method `#newKeyring` removed a data validation because of the `data` param input from the clients.
  - The method `fullUpdate` was mistakenly flagged as private, this PR introduces it again as public.

## References

- https://github.com/MetaMask/KeyringController/pull/202
- https://github.com/MetaMask/KeyringController/pull/221

## Checklist


gantunesr:
@legobeat good point!

> _How can I replicate the release-blocking issue that this addresses locally in order to review the fix?_

I'm not sure about the exact problem that led to the error in NPM publishing. I was reviewing the previous workflow but didn't find any logs pertaining to the issue. Previously, I encountered errors concerning the `yarn build` command and encountered some typescript problems with the clients. @Gudahtt, do you have any suggestions on how we can ensure that the publish workflow is functioning correctly?

paulmillr:
any updates on this one?

gantunesr:
@paulmillr I'll run a test now and if it succeeds it should be good to go

paulmillr:
@gantunesr tests seems to pass

gantunesr:
@paulmillr I was referring to the unit and E2E tests in the MetaMask client, they're failing due to the error management updates. I'm working on fixing them

jest.config.js
```diff
@@ -41,10 +41,10 @@ module.exports = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 68.67,
-      functions: 92.59,
-      lines: 90.2,
-      statements: 90.42,
+      branches: 70.37,
```
owencraston, line 44: is there any reasoning behind these numbers?

HowardBraham: I believe there are semi-automatic rules that bump them up sometimes

src/KeyringController.ts
```diff
@@ -443,13 +443,13 @@ class KeyringController extends EventEmitter {
...
   ): Promise<string> {
-    const address = normalizeToHex(msgParams.from);
+    const address = normalizeToHex(msgParams.from) as Hex;
     const keyring = await this.getKeyringForAccount(address);
     if (!keyring.signPersonalMessage) {
       throw new Error(KeyringControllerError.UnsupportedSignPersonalMessage);
     }
 
-    const normalizedData = normalizeToHex(msgParams.data);
+    const normalizedData = normalizeToHex(msgParams.data) as Hex;
```
owencraston, line 452: why not have the normalizeToHex return the proper Hex value you are expecting?

jest.config.js
gantunesr: Correct, it indicates the code coverage and it gets updated every time you run the tests

src/KeyringController.ts
gantunesr: It's part of another module I have to update, this is just a temporary measure

owencraston (approved):

paulmillr:
Is timing for v11 release known?

PR #234, opened 2023-06-21
Update utils package and rename dispose fn

mikesposito:
## Description

This PR pulls the latest version of `@metamask/utils`, which also includes an updated `Keyring` type with an optional `destroy` method: this allows us to remove some type narrowing and the ambiguity between `dispose` and `destroy` methods. 

## Changes

- **CHANGED**: `dispose` optional method is not called anymore when destroying a keyring

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@metamask/eth-hd-keyring": "^6.0.0",
     "@metamask/eth-sig-util": "^6.0.0",
     "@metamask/eth-simple-keyring": "^5.0.0",
-    "@metamask/utils": "^5.0.0",
+    "@metamask/utils": "^6.1.0",
     "obs-store": "^4.0.3"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


mcmire (approved):
Looks good!

Gudahtt (changes requested):
I think we need to bump the minimum Node.js version first, this package still supports v14

package.json
```diff
@@ -81,7 +81,7 @@
   },
   "packageManager": "yarn@3.2.1",
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
```
Gudahtt, line 84: Thoughts on moving this to a separate PR? It would be easier to keep track of that way

mikesposito: Yes, I think it would be better to separate the scopes into two PRs

I created [236](https://github.com/MetaMask/KeyringController/pull/236), that we could merge before this PR

Gudahtt (approved):
LGTM!

PR #235, opened 2023-07-04
Update `@metamask/eth-sig-util` to `^6`

mikesposito:
## Description

This PR updates `@metamask/eth-sig-util` package to `^6`, which includes a change on the `normalize` function that affects how empty strings and `0` inputs are treated, and since `@metamask/eth-keyring-controller` normalizes data passed to `signPersonalMessage` since v11, I think this change should be breaking also for `@metamask/eth-keyring-controller`.

## Changes

- **BREAKING**: `signPersonalMessage` now normalizes `msgParams.data` in a different way for `0` and empty strings inputs. 
    - `0` will be normalized to `0x00` and empty strings to `0x` 

## References

* Related to [#1315](https://github.com/MetaMask/core/issues/1315)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "@ethereumjs/tx": "^4.1.1",
     "@metamask/browser-passworder": "^4.1.0",
     "@metamask/eth-hd-keyring": "^6.0.0",
-    "@metamask/eth-sig-util": "^5.1.0",
+    "@metamask/eth-sig-util": "^6.0.0",
     "@metamask/eth-simple-keyring": "^5.0.0",
     "@metamask/utils": "^5.0.0",
     "obs-store": "^4.0.3"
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

Gudahtt:
Though... is this breaking? I can see that it's called a lot with the `address` parameter, but that's intended to be an address. I'm not sure any of these methods worked in the first place if an empty string or zero was passed in. And certainly MetaMask would never pass in either of those, we don't let RPC calls get to the keyring until after verifying the address.

I guess it's breaking only due to a lack of expressive types/validation prior to this, but not in a way that impacts us.

The only exception to that is for the personal sign method. In that case, we use this normalize function for the data as well. Again the expectation seems clear that it's meant to be a hex string, but this might affect "empty" personal sign signatures potentially.

PR #238, opened 2023-07-05
fix: throw correct error on empty or invalid address

mikesposito:
## Description

Since `@metamask/utils` has been updated to the latest version, `normalize` method does not return `undefined` for empty strings anymore. 

`getKeyringForAccount` checks if the provided address is an empty string (or undefined) and returns an error in case. But as all the other methods of the controller normalize the address before calling `getKeyringForAccount` what we check in there is `0x`, which is not seen as invalid/empty by the function.

To fix the problem and provide a correct error message, this PR uses `isValidHexAddress` from `@metamask/utils` to check if the string passed to `getKeyringForAccount` is a valid address.

## Changes

- **FIXED**: `getKeyringForAccount` provides a correct error message when an address is invalid

## References

* Relates to [#1315](https://github.com/MetaMask/core/issues/1315)

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -43,8 +43,8 @@ module.exports = {
     global: {
       branches: 72.09,
       functions: 92.85,
-      lines: 90.84,
-      statements: 91.05,
+      lines: 90.81,
+      statements: 91.02,
     },
   },
   preset: 'ts-jest',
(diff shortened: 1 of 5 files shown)


mikesposito:
Could be worth waiting for https://github.com/MetaMask/utils/pull/112 to be merged and released as  `@metamask/controller-utils` will not be needed anymore

mikesposito:
> LGTM but can we add a test-case for at least one empty/falsey case (ie covering for when the previous `!address` on line 719 would hold)?

Shouldn't it be still covered by the test case `throws error when address is not provided (L715)`?

PR #247, opened 2023-07-17
Add types for store and fix type discrepancies

mikesposito:
## Description

This PR updates `obs-store` to `@metamask/obs-store` and fixes some type discrepancies with the actual states

## Changes

- **BREAKING**: constructor opts `KeyringControllerArgs` properties types changed
  - Simplified type for `keyringBuilders`, now is an optional `{ (): Keyring<Json>; type: string }[]`
  - `initState` now accepts only an optional persistent state, of type `KeyringControllerPersistentState`
- **BREAKING**: type of `store` and `memStore` public properties changed
  - `KeyringController.store` is now an `ObservableStore<KeyringControllerPersistentState>`
  - `KeyringController.memStore` is now an `ObservableStore<KeyringControllerState>`
- **BREAKING**: `updateMemStoreKeyrings` method return type changed to `Promise<void>`
- **BREAKING**: `KeyringControllerState` type changed to include only non-persistent state:
  - Now `undefined` is used instead of `null` when `encryptionKey` and `encryptionSalt` are unset 
  - `keyrings` is now of type `KeyringObject[]` instead of `Keyring<Json>` 
  - `password`, `store`, `memStore` have been removed - note that in practice this change only affects types
  - This changes cause the following methods also to change the return type:
    - `createNewVaultAndKeychain`
    - `createNewVaultAndRestore`
    - `setLocked`
    - `submitPassword`
    - `submitEncryptionKey`
    - `addNewAccount`
    - `removeAccount`
    - `fullUpdate`
- **ADDED**: `KeyringControllerPersistentState` type which includes only persistent state, an optional string property with key `vault`
- **ADDED**: `KeyringObject` type for how keyrings are represented in `memStore`

## Checklist

### DIFF
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -122,7 +122,7 @@ describe('KeyringController', () => {
       await keyringController.persistAllKeyrings();
 
       const { vault } = keyringController.store.getState();
-      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault);
+      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault as string);
       expect(keyrings).toContain(unsupportedKeyring);
       expect(keyrings).toHaveLength(2);
     });
@@ -136,7 +136,9 @@ describe('KeyringController', () => {
         const vault = JSON.stringify({ salt: vaultEncryptionSalt });
         keyringController.store.updateState({ vault });
 
-        expect(keyringController.memStore.getState().encryptionKey).toBeNull();
+        expect(
+          keyringController.memStore.getState().encryptionKey,
+        ).toBeUndefined();
         expect(
           keyringController.memStore.getState().encryptionSalt,
         ).toBeUndefined();
@@ -201,7 +203,7 @@ describe('KeyringController', () => {
 
   describe('createNewVaultAndKeychain', () => {
     it('should create a new vault', async () => {
-      keyringController.store.updateState({ vault: null });
+      keyringController.store.putState({});
       assert(!keyringController.store.getState().vault, 'no previous vault');
 
       const newVault = await keyringController.createNewVaultAndKeychain(
@@ -213,7 +215,7 @@ describe('KeyringController', () => {
     });
 
     it('should unlock the vault', async () => {
-      keyringController.store.updateState({ vault: null });
+      keyringController.store.putState({});
       assert(!keyringController.store.getState().vault, 'no previous vault');
 
       await keyringController.createNewVaultAndKeychain(PASSWORD);
@@ -222,7 +224,7 @@ describe('KeyringController', () => {
     });
 
     it('should encrypt keyrings with the correct password each time they are persisted', async () => {
-      keyringController.store.updateState({ vault: null });
+      keyringController.store.putState({});
       assert(!keyringController.store.getState().vault, 'no previous vault');
 
       await keyringController.createNewVaultAndKeychain(PASSWORD);
@@ -254,7 +256,7 @@ describe('KeyringController', () => {
         await keyringController.createNewVaultAndKeychain(PASSWORD);
         const finalMemStore = keyringController.memStore.getState();
 
-        expect(initialMemStore.encryptionKey).toBeNull();
+        expect(initialMemStore.encryptionKey).toBeUndefined();
         expect(initialMemStore.encryptionSalt).toBeUndefined();
 
         expect(finalMemStore.encryptionKey).toBe(MOCK_HARDCODED_KEY);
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -1,32 +1,29 @@
 import type { Json, Keyring } from '@metamask/utils';
-import ObservableStore from 'obs-store';
 
 export type KeyringControllerArgs = {
   keyringBuilders:
     | { (): Keyring<Json>; type: string }
     | ConcatArray<{ (): Keyring<Json>; type: string }>;
```
Gudahtt, line 6: Perhaps this `ConcatArray<{ (): Keyring<Json>; type: string }>` part can be removed as well? I'm not sure why this is here. It seems that `keyringBuilders` is expected to be a regular array.

src/types.ts
```diff
@@ -1,32 +1,26 @@
 import type { Json, Keyring } from '@metamask/utils';
-import ObservableStore from 'obs-store';
 
 export type KeyringControllerArgs = {
-  keyringBuilders:
-    | { (): Keyring<Json>; type: string }
-    | ConcatArray<{ (): Keyring<Json>; type: string }>;
-
+  keyringBuilders: { (): Keyring<Json>; type: string }[];
   cacheEncryptionKey: boolean;
```
Gudahtt, line 5: Should we mark this as optional as well? Since it was effectively optional in older versions

Gudahtt (commented):
Looks great! Let me know when the description is ready so that I can review the list of changes. Lots of changes here.

Gudahtt:
@SocketSecurity ignore @metamask/obs-store@8.1.0
    @SocketSecurity ignore @metamask/safe-event-emitter@2.0.0

These are our packages

mikesposito:
The description should now be ready, I hope I mentioned everything 😄

src/KeyringController.ts
```diff
@@ -167,15 +167,15 @@ class KeyringController extends EventEmitter {
   async setLocked(): Promise<KeyringControllerState> {
     delete this.password;
 
+    // remove keyrings
```
mcmire, line 170: Why was this change necessary? Just curious.

Gudahtt:
I have created a few extra PRs for the problems I found related to `signTypedMessage`:

* https://github.com/MetaMask/KeyringController/pull/248
* https://github.com/MetaMask/KeyringController/pull/249
* https://github.com/MetaMask/KeyringController/pull/250

src/KeyringController.test.ts
```diff
@@ -122,7 +122,7 @@ describe('KeyringController', () => {
       await keyringController.persistAllKeyrings();
 
       const { vault } = keyringController.store.getState();
-      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault);
+      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault as string);
```
mcmire, line 125: Ah, I see that the `vault` is not set initially, so there's no way to remove the `undefined` from its type.

This comes up in the NetworkController tests (`provider` is similarly not set initially and set is later on) and the way I've solved is to assert it in the test if I need to do something with it. So what you do think about something like:

``` suggestion
      assert(vault, "Vault is not set");
      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault);
```

mcmire (commented):
Did a quick pass on this and had a couple of suggestions/questions. Will do another pass tomorrow to make sure I understand how these types are changing.

src/KeyringController.ts
mikesposito: It wasn't indeed, I believed `#clearKeyrings` looped over keyrings in the store, but they are an array of instances on the class property.
I rolled back the change

src/KeyringController.ts
```diff
@@ -898,10 +897,12 @@ class KeyringController extends EventEmitter {
...
-        this.memStore.updateState({
-          encryptionKey,
-          encryptionSalt,
-        });
+        this.memStore.putState(
+          Object.assign(this.memStore.getState(), {
+            encryptionKey,
+            encryptionSalt,
+          }),
+        );
```
Gudahtt, line 905: Nit: Is this change necessary? This seems equivalent to what was here before

```suggestion
        this.memStore.updateState({
          encryptionKey,
          encryptionSalt,
        });
```

mikesposito: I think it would raise a type error because encryptionKey and encryptionSalt could be undefined, which doesn't mean that the two keys will be removed from the state

mikesposito: With `putState` and an object assign we pass the current state and encryptionKey and encryptionSalt _if they are defined_

Gudahtt: It should already be impossible for them to be `undefined` here. This block is when `cacheEncryptionKey` is set. A few lines above here we already verify that `encryptionKey` is a string and `encryptionSalt` matches the salt in the encrypted vault

mikesposito: I rolled back, adding a non-null assertion on the `encryptionSalt`

Gudahtt (approved):
LGTM! Description of changes looks great too

PR #248, opened 2023-07-17
Fix `@metamask/eth-sig-util` types

Gudahtt:
## Description

The types for the package `@metamask/eth-sig-util` were being overwritten by a local type declaration. That has been removed, and all type errors have been resolved.

All of the type errors were related to `normalize` returning `undefined` if given a nullish input. We never pass it a nullish input, but TypeScript wasn't able to infer that, so that issue was addressed with type casts. This isn't ideal, but it's temporary. Once we finish migrating all Keyrings to the new Keyring interface, we won't need to normalize addresses anymore at all.

## Changes

None

## References

I was attempting to fix type errors in this package that were discovered while reviewing https://github.com/MetaMask/core/pull/1441. I was unable to use additional types from `@metamask/eth-sig-util` due to the problem fixed by this PR.

## Checklist

### DIFF
--- a/src/KeyringController.ts
+++ b/src/KeyringController.ts
@@ -334,7 +334,9 @@ class KeyringController extends EventEmitter {
       return res.concat(arr);
     }, []);
 
-    return addresses.map(normalizeToHex);
+    // Cast to `Hex[]` here is safe here because `addresses` has no nullish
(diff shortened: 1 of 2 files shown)


mikesposito (approved):
LGTM!

PR #249, opened 2023-07-17
Narrow return type of `signTypedMessage` and encryption methods

Gudahtt:
## Description

The methods `signTypedMessage`, `getEncryptionPublicKey`, and `decryptMessage` always return a string, but the declared return type was `Bytes` (which is a type union of `string` along with other things). This made the methods more difficult to use, as callers had to handle the other `Bytes` types as return values as well. 

The methods have been updated to use `string` as the return type.

## Changes

- Changed: The methods `signTypedMessage`, `getEncryptionPublicKey`, and `decryptMessage`  now return `string` rather than `Bytes`
  - This is non-breaking because the `Bytes` type includes `string`

## References

Helps to unblock https://github.com/MetaMask/core/pull/1441

## Checklist

### DIFF
--- a/src/KeyringController.ts
+++ b/src/KeyringController.ts
@@ -7,7 +7,6 @@ import { remove0x, isValidHexAddress } from '@metamask/utils';
 import type {
   Hex,
   Json,
-  Bytes,
   Keyring,
   KeyringClass,
   Eip1024EncryptedData,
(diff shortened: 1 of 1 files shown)


mikesposito (approved):
LGTM!

I just noticed we have the same problem with `getEncryptionPublicKey` and`decryptMessage`, we can fix them separately but we could also take this chance

Gudahtt:
Oh, good catch! I have updated those two methods as well, and updated the PR title & description accordingly.

Figured we may as well fix them here since it's the same problem.

mikesposito (approved):

PR #250, opened 2023-07-17
Fix `signTypedMessage` parameter types

Gudahtt:
## Description

The `signTypedMessage` parameter types accepted the data to sign as an array. This works with `signTypedData_v1`, but is incompatible with later versions.

The signature has been updated to accept data as either an array or an object. Additionally, tests have been added to cover all three supported versions of `signTypedData`.

## Changes

- Fixed: `signTypedMessage` now accepts data either as an array or an object
  - This was a mistake in the original type definition - it always supported both in practice.

## References

Helps to unblock https://github.com/MetaMask/core/pull/1441

## Checklist


mikesposito (approved):
LGTM!

mikesposito (approved):

PR #251, opened 2023-07-17
Refactor controller initialization in tests

Gudahtt:
## Description

Each test now uses a helper function to construct the keyring controller rather than using `beforeEach`. This eliminates redundant redundancy in cases where the controller was being overwritten, and makes it easier for future tests to customize how the controller is initialized.

## Changes

None

## References

N/A

## Checklist

### DIFF
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -14,6 +14,7 @@ import {
   MOCK_HARDCODED_KEY,
   MOCK_HEX,
 } from './test';
+import type { KeyringControllerArgs } from './types';
 
 const MOCK_ENCRYPTION_KEY =
   '{"alg":"A256GCM","ext":true,"k":"wYmxkxOOFBDP6F6VuuYFcRt_Po-tSLFHCWVolsHs4VI","key_ops":["encrypt","decrypt"],"kty":"oct"}';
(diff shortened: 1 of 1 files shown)


src/KeyringController.test.ts
```diff
@@ -37,26 +37,59 @@ const walletTwoAddresses = [
...
+  if (seedPhrase && !password) {
+    throw new Error('Password required to restore vault');
+  } else if (seedPhrase && password) {
+    await keyringController.createNewVaultAndRestore(
+      PASSWORD,
+      walletOneSeedWords,
+    );
+  } else if (password) {
     await keyringController.createNewVaultAndKeychain(PASSWORD);
     await keyringController.submitPassword(PASSWORD);
```
mikesposito, line 77: Nit: this should not be needed as unlock is already done inside `createNewVaultAndKeychain`

mikesposito (approved):
Looks good, I just left a couple of notes

src/KeyringController.test.ts
Gudahtt: Huh, good point. I was just copying this from the pre-existing `beforeEach`. But yes we can remove this. Done in 7195eb1

src/KeyringController.test.ts
```diff
@@ -37,26 +37,59 @@ const walletTwoAddresses = [
...
+ * @returns A keyring controller.
+ */
+async function initializeKeyringController({
+  constructorOptions,
+  password,
+  seedPhrase,
+}: {
+  constructorOptions?: Partial<
+    ConstructorParameters<typeof KeyringController>[0]
+  >;
```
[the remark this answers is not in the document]
Gudahtt: Good idea - done in 53ab639

mikesposito (approved):

PR #252, opened 2023-07-18
Restore compatibility with QR Keyring

Gudahtt:
## Description

The `KeyringController` method `addNewKeyring` was changed in v11 to initialize the keyring options to an empty object. These options are passed to the `deserialize` method of the keyring. This broke compatibility with the QR keyring because its `deserialize` method does not expect to be passed an empty object.

The `KeyringController` has been updated to no longer initialize the `addNewKeyring` options to an empty object. It should now work the same as it did prior to v11. The one functional difference is that it now throws an error when `addNewKeyring` is used to add a simple keyring without supplying any private keys, however that path would have already thrown an error.

## Changes

- Fixed: Fix `addNewKeyring` incompatbility when no options are provided
  - The `addNewKeyring` method stopped working with certain keyrings (e.g. the QR keyring) in v11 when called with no options. 

## References

This bug was introduced before and fixed in the past, e.g. https://github.com/MetaMask/KeyringController/pull/136

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -41,10 +41,10 @@ module.exports = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 72.09,
+      branches: 72.41,
       functions: 92.85,
-      lines: 90.81,
-      statements: 91.02,
+      lines: 90.87,
+      statements: 91.08,
     },
   },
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon #251

mikesposito (approved):
LGTM!

PR #253, opened 2023-07-18
Allow non-object keyring serialized state

Gudahtt:
## Description

The `addNewKeyring` method has been updated to support non-object keyring serialized state, which had been supported already on versions prior to v11.

This allows us to remove a special-case for the simple keyring, which requires a JSON array as its serialized keyring state.

When constructing a simple keyring, the method `addNewKeying` now expects the second parameter (`opts`) to be an array of private keys rather than an object with a `privateKeys` property. This is how it worked prior to v11.

## Changes

- **BREAKING**: When constructing a simple keyring with `addNewKeying`, the second parameter (`opts`) is now expected to be an array of private keys rather than an object with a `privateKeys` property
  - This is how it worked prior to v11
- Changed: Restored support for keyrings with non-object serialized state
  - This was supported prior to v11

## References

None

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -41,10 +41,10 @@ module.exports = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 72.41,
+      branches: 71.42,
       functions: 92.85,
-      lines: 90.87,
-      statements: 91.08,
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon #252

PR #262, opened 2023-09-01
devDeps: bump `@metamask/eslint-config*` to v12

legobeat:
## Description

Combines:
- #246 
- #256 
- #257 
- #270
- aligns `peerDependencies`
- bumps devDeps
- applies new linting

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@lavamoat/allow-scripts": "^2.3.1",
     "@lavamoat/preinstall-always-fail": "^1.0.0",
     "@metamask/auto-changelog": "^3.0.0",
-    "@metamask/eslint-config": "^12.1.0",
+    "@metamask/eslint-config": "^12.2.0",
     "@metamask/eslint-config-jest": "^12.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,6 +38,10 @@
     "test": "jest && jest-it-up",
     "test:watch": "jest --watch"
   },
+  "resolutions": {
+    "ink": "3.1.0",
+    "react-reconciler": "0.24.0"
+  },
```
jiexi, line 44: why are these being forced?

jiexi: `ink` 3.2.0 has peer deps for `"react": ">=16.8.0"` and only devDeps for react 17. So it should be fine??

I see why `react-reconciler` must be in the resolutions though

jiexi: Doesn't `"react-reconciler": "0.24.0"` already force the version of `react-reconcile` in `ink`? No big deal, just trying to understand if I had a wrong assumption of how the resolution works in yarn. Going to approve

jiexi (approved):

jiexi (approved):

PR #293, opened 2023-11-07
Add `encryptor` types

### LINKED ISSUE #295: `encryptor` lacks sufficient typing
Currently, the `encryptor` class variable (and the corresponding constructor option) is typed with `any`.
As a controller client can inject its own encryptor methods, the minimum encryptor interface needed by the controller should be clear.  

Moreover, the solution should take into account that:
- MetaMask extension and mobile use different encryptors
  - MetaMask extension uses the default encryptor value, provided by KeyringController, which is `@metamask/browser-passworder`
  - MetaMask mobile injects its own encryptor.
- When `cacheEncryptionKey` options is set to `true`, `KeyringController` needs an encryptor capable of exporting the encryption key string along with encrypted/decrypted strings. This is only needed by the extension, when using MV3
  - This means that when `cacheEncryptionKey` is set to `true`, the encryptor **must** implement additional methods specific for this

mikesposito:
## Description

Currently `encryptor` type is `any`, and as different clients can use different encryptor utilities, we need to establish a minimum supported (or needed) interface that the encryptor should implement.

This PR adds some types and guards to the encryptor, taking into account `@metamask/browser-passworder`, which is the encryptor used by the extension, and the [Encryptor class](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Encryptor.js) used by mobile

Restrictions have also been added for `cacheEncryptionKey` and `encryptor`, as the methods that the encryptor must support depends on whether we need the encryption key to be exported or not. To do this, public class variables have been substituted with sharp vars and a check on construction has been added: an error is thrown when they are incompatible 

## Changes

- **BREAKING**: `encryptor` class variable is now inaccessible
- **BREAKING**: `cacheEncryptionKey` class variable is now inaccessible
- **BREAKING**: `encryptor` constructor option type has been changed to `GenericEncryptor | ExportableKeyEncryptor | undefined`

## References

Waiting on https://github.com/MetaMask/browser-passworder/pull/49
* Fixes #295 

## Checklist

### DIFF
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -15,7 +15,7 @@ import {
   MOCK_HARDCODED_KEY,
   MOCK_ENCRYPTION_SALT,
 } from './test';
-import type { KeyEncryptor, KeyringControllerArgs } from './types';
+import type { ExportableKeyEncryptor, KeyringControllerArgs } from './types';
 
 const MOCK_ENCRYPTION_KEY =
   '{"alg":"A256GCM","ext":true,"k":"wYmxkxOOFBDP6F6VuuYFcRt_Po-tSLFHCWVolsHs4VI","key_ops":["encrypt","decrypt"],"kty":"oct"}';
@@ -856,7 +856,7 @@ describe('KeyringController', () => {
           );
           const updateVaultStub = sinon
             .stub(
-              keyringController.encryptor as KeyEncryptor,
+              keyringController.encryptor as ExportableKeyEncryptor,
               'updateVaultWithDetail',
             )
             .resolves({
@@ -1091,10 +1091,10 @@ describe('KeyringController', () => {
         },
       });
       const returnValue = await (
-        keyringController.encryptor as KeyEncryptor
+        keyringController.encryptor as ExportableKeyEncryptor
       ).decryptWithKey('', '');
       const decryptWithKeyStub = sinon.stub(
-        keyringController.encryptor as KeyEncryptor,
+        keyringController.encryptor as ExportableKeyEncryptor,
         'decryptWithKey',
       );
       decryptWithKeyStub.resolves(Promise.resolve(returnValue));
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -88,6 +88,7 @@ describe('KeyringController', () => {
       it('should throw error if provided encryptor does not support key export', async () => {
         expect(
           () =>
+            // @ts-expect-error we want to bypass typechecks here.
             new KeyringController({
               cacheEncryptionKey: true,
               encryptor: {
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -9,17 +9,16 @@ import * as sinon from 'sinon';
 import { KeyringController, keyringBuilderFactory } from '.';
 import { KeyringType, KeyringControllerError } from './constants';
 import {
-  mockEncryptor,
+  MockEncryptor,
   KeyringMockWithInit,
   PASSWORD,
   MOCK_HARDCODED_KEY,
-  MOCK_HEX,
+  MOCK_ENCRYPTION_SALT,
 } from './test';
-import type { KeyringControllerArgs } from './types';
+import type { KeyEncryptor, KeyringControllerArgs } from './types';
 
 const MOCK_ENCRYPTION_KEY =
   '{"alg":"A256GCM","ext":true,"k":"wYmxkxOOFBDP6F6VuuYFcRt_Po-tSLFHCWVolsHs4VI","key_ops":["encrypt","decrypt"],"kty":"oct"}';
-const MOCK_ENCRYPTION_SALT = 'HQ5sfhsb8XAQRJtD+UqcImT7Ve4n3YMagrh05YTOsjk=';
 const MOCK_ENCRYPTION_DATA = `{"data":"2fOOPRKClNrisB+tmqIcETyZvDuL2iIR1Hr1nO7XZHyMqVY1cDBetw2gY5C+cIo1qkpyv3bPp+4buUjp38VBsjbijM0F/FLOqWbcuKM9h9X0uwxsgsZ96uwcIf5I46NiMgoFlhppTTMZT0Nkocz+SnvHM0IgLsFan7JqBU++vSJvx2M1PDljZSunOsqyyL+DKmbYmM4umbouKV42dipUwrCvrQJmpiUZrSkpMJrPJk9ufDQO4CyIVo0qry3aNRdYFJ6rgSyq/k6rXMwGExCMHn8UlhNnAMuMKWPWR/ymK1bzNcNs4VU14iVjEXOZGPvD9cvqVe/VtcnIba6axNEEB4HWDOCdrDh5YNWwMlQVL7vSB2yOhPZByGhnEOloYsj2E5KEb9jFGskt7EKDEYNofr6t83G0c+B72VGYZeCvgtzXzgPwzIbhTtKkP+gdBmt2JNSYrTjLypT0q+v4C9BN1xWTxPmX6TTt0NzkI9pJxgN1VQAfSU9CyWTVpd4CBkgom2cSBsxZ2MNbdKF+qSWz3fQcmJ55hxM0EGJSt9+8eQOTuoJlBapRk4wdZKHR2jdKzPjSF2MAmyVD2kU51IKa/cVsckRFEes+m7dKyHRvlNwgT78W9tBDdZb5PSlfbZXnv8z5q1KtAj2lM2ogJ7brHBdevl4FISdTkObpwcUMcvACOOO0dj6CSYjSKr0ZJ2RLChVruZyPDxEhKGb/8Kv8trLOR3mck/et6d050/NugezycNk4nnzu5iP90gPbSzaqdZI=","iv":"qTGO1afGv3waHN9KoW34Eg==","salt":"${MOCK_ENCRYPTION_SALT}"}`;
 
 const walletOneSeedWords =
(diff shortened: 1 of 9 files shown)


src/KeyringController.ts
```diff
@@ -36,16 +43,16 @@ class KeyringController extends EventEmitter {
...
 
-  public cacheEncryptionKey: boolean;
-
   public unsupportedKeyrings: SerializedKeyring[];
 
   public password?: string;
 
+  #encryptor: GenericEncryptor | KeyEncryptor;
+
+  #cacheEncryptionKey: boolean;
```
mikesposito, line 54: These has been made unaccessible, so that we can add a getter and a setter for each of them, and add validations on setters

src/KeyringController.ts
```diff
@@ -907,11 +925,63 @@ class KeyringController extends EventEmitter {
...
+   * Setter for the encryptor.
+   *
+   * @param encryptor - The encryptor to set.
+   * @throws If the `cacheEncryptionKey` option is true and the
+   * encryptor is not a valid KeyEncryptor.
+   */
+  set encryptor(encryptor: GenericEncryptor | KeyEncryptor) {
+    if (this.cacheEncryptionKey) {
+      assertIsKeyEncryptor(encryptor);
+    }
```
mikesposito, line 947: We can't cache the encryption key if the provided encryptor does not support it

src/test/encryptor.mock.ts
mikesposito, line 1: This takes inspiration from `@metamask/keyring-controller`: https://github.com/MetaMask/core/blob/main/packages/keyring-controller/tests/mocks/mockEncryptor.ts

src/KeyringController.ts
Gudahtt: Have you considered making this fully private? We don't have any reason to access the encyrptor at runtime, or to change it at runtime. The constructor parameter seems adequate to me, and reduces the API surface area

src/KeyringController.ts
```diff
@@ -1053,6 +1123,66 @@ class KeyringController extends EventEmitter {
...
+
+  /**
+   * Ensure that the vault is encrypted with the latest encryption method available.
+   *
+   * The vault is left unchanged if no password is available, if the encryptor does not support
+   * the vault update, or if the vault is already encrypted with the latest encryption method.
+   *
+   * @returns The updated encrypted vault.
+   */
+  async #updateVaultEncryption(): Promise<void> {
```
Gudahtt, line 1135: Thoughts on moving this to the extension instead? Not sure what advantage it has at this layer. The clients are already use separate encryptors and separate encryption configuration.

Moving this to the client would also make it easier to extract encryption from the keyring controller altogether. That's something the accounts team was working on, to support encrypting more types of data, not just keyrings.

mikesposito: As `eth-keyring-controller` instance is unaccessible from the extension, we could update the `keyring-controller` store (and thus also the extension storage), but as `eth-keyring-controller` does not listen to `keyring-controller` state updates (it is only the other way around) the two controllers would then be in an inconsistent state. I'm not sure this would represent a problem though - I have to look better into that, but this might need some change in `keyring-controller` in case we want to handle this in the extension

mikesposito: Perhaps the extension can use the encryptor to update the vault in the storage prior to keyring controller initialisation

Gudahtt: Hmm, I was thinking this was an operation on the encryptor itself, but it's not. I see what you mean.

mikesposito: Yes, exactly. Indeed, this is only needed to be sure that the vault gets migrated even if the user doesn't make any action on the keychain

Gudahtt: Gotcha, makes sense. Instead of having this custom migrate function then, could we accomplish this by calling `persistAllKeyrings()` in `unlockKeyrings`?

mikesposito: We could also call `persistAllKeyrings`, but:
- it should be done after the unlock or we'll overwrite the existing vault with an empty one
- EthKeyringController would have no way of knowing if the vault is already up to date, so it would repeat the operation every time we unlock the vault, which is not great

mikesposito: In https://github.com/MetaMask/KeyringController/pull/293/commits/29713e605d2009ebd3a62f66c7ee829d9db96402 we now:
- Execute the whole `unlockKeyrings`, so that the keyrings array is up to date with the saved vault
- Check the current vault against the one returned by `updateVault`
- Call `persistAllKeyrings` in case they differ

src/KeyringController.test.ts
```diff
@@ -382,16 +345,14 @@ describe('KeyringController', () => {
...
+          },
         });
-        keyringController.cacheEncryptionKey = true;
 
-        const initialMemStore = keyringController.memStore.getState();
         await keyringController.createNewVaultAndKeychain(PASSWORD);
-        const finalMemStore = keyringController.memStore.getState();
-
-        expect(initialMemStore.encryptionKey).toBeUndefined();
-        expect(initialMemStore.encryptionSalt).toBeUndefined();
```
mikesposito, line 332: These are not applicable anymore, as `cacheEncryptionKey` is set to true on construction (before the first `createNewVaultAndKeychain > persistAllKeyrings` is called), so `encryptionKey` and `encryptionSalt` will be already defined

src/KeyringController.ts
mikesposito: Good point, I agree! Refactored in aff13b04678d025059c6e3ccf3c6a1ca989e358f
This change also required some tests refactoring

src/KeyringController.test.ts
```diff
@@ -751,18 +771,121 @@ describe('KeyringController', () => {
...
+            .stub(mockEncryptor, 'updateVault')
+            .resolves(updatedVault);
+          const keyringController = await initializeKeyringController({
+            password: PASSWORD,
+            constructorOptions: {
+              cacheEncryptionKey: true,
+              encryptor: mockEncryptor,
+            },
+          });
+          const persistAllKeyringsSpy = sinon.spy(
```
Gudahtt, line 853: Nit: Effectively we're testing an internal implementation here by using a spy to check that an internal call is made. Perhaps we could test the effects instead by looking at the store both before and after the keyrings are unlocked?

src/KeyringController.ts
Gudahtt: Looks great!

Gudahtt (approved):
A few pending suggestions/questions but overall this LGTM!

src/KeyringController.ts
Gudahtt: Looks great!

src/KeyringController.test.ts
```diff
@@ -161,39 +178,39 @@ describe('KeyringController', () => {
...
 
-      const { vault } = keyringController.store.getState();
-      assert(vault, 'Vault is not set');
-      const keyrings = await mockEncryptor.decrypt(PASSWORD, vault);
-      expect(keyrings).toContain(unsupportedKeyring);
-      expect(keyrings).toHaveLength(2);
+      assert(keyringController.store.getState().vault, 'Vault is not set');
+      expect(encryptSpy.calledOnce).toBe(true);
+      expect(encryptSpy.getCalls()?.[0]?.args[1]).toHaveLength(2);
+      expect(encryptSpy.getCalls()?.[0]?.args[1]).toContain(unsupportedKeyring);
```
mcmire, line 197: Nit: Is the first `?.` necessary here?

``` suggestion
      expect(encryptSpy.getCalls()[0]?.args[1]).toHaveLength(2);
      expect(encryptSpy.getCalls()[0]?.args[1]).toContain(unsupportedKeyring);

**Review by mcmire — DISMISSED** [review]
I had some nits as well, but this makes sense and looks good.

File: src/KeyringController.ts [hunk 7]
```diff
@@ -789,30 +799,34 @@ class KeyringController extends EventEmitter {
...
       } else if (encryptionKey) {
-        const key = await this.encryptor.importKey(encryptionKey);
-        const vaultJSON = await this.encryptor.encryptWithKey(
+        const key = await this.#encryptor.importKey(encryptionKey);
+        const vaultJSON = await this.#encryptor.encryptWithKey(
           key,
           serializedKeyrings,
         );
-        vaultJSON.salt = encryptionSalt;
+        if (encryptionSalt) {
```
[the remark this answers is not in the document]
mikesposito: I changed the `KeyringControllerState` type to:

```typescript
export type KeyringControllerState = {
  keyrings: KeyringObject[];
  isUnlocked: boolean;
  encryptionKey?: string;
  encryptionSalt?: string;
} & (
  | { encryptionKey: string; encryptionSalt: string }
  | {
      encryptionKey?: never;
      encryptionSalt?: never;
    }
);
```

Is there a better way to represent this?

src/KeyringController.test.ts
mikesposito: Good point. I refactored this in f2382456a7a4298ffd9df83484ef0fc7345ef23c

src/KeyringController.ts
```diff
@@ -903,12 +919,27 @@ class KeyringController extends EventEmitter {
...
+      throw new Error(KeyringControllerError.VaultDataError);
+    }
+
     await Promise.all(vault.map(this.#restoreKeyring.bind(this)));
     await this.updateMemStoreKeyrings();
+
+    if (
+      this.password &&
+      this.#encryptor.updateVault &&
+      (await this.#encryptor.updateVault(encryptedVault, this.password)) !==
```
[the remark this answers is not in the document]
mikesposito: Yes, that's quite bad. The best way would be to expose `isVaultUpdated` from `browser-passworder`, because I don't think there's a nice way to inline that logic here, as `eth-keyring-controller` would have to know the latest vault parameters (`DEFAULT_DERIVATION_PARAMS` in `browser-passworder`, which is also not exposed)

Gudahtt (approved):
LGTM!

src/KeyringController.ts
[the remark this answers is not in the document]
Gudahtt: Nice! Looks reasonable

PR #307, opened 2023-11-23
Prefer cached `encryptionKey` for encryption when possible

mikesposito:
## Description

When persisting keyrings, `EthKeyringController` always prefers the use of the password over the cached encryption key. This means that we'll always derive the password (which is an heavy task), even if we already have the key that we need.

This PR changes the order of the check, putting the `encryptionKey` on top. Some changes were required on some tests after that as well. 

## Changes

- **FIXED**: Use cached encryption key when available

## References

- See https://github.com/MetaMask/metamask-extension/pull/21878#issuecomment-1824568343

## Checklist

### DIFF
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -887,6 +887,35 @@ describe('KeyringController', () => {
         });
       });
 
+      describe(`with cacheEncryptionKey = true and encryptionKey is set`, () => {
+        it('should not update the vault', async () => {
+          const mockEncryptor = new MockEncryptor();
+          const keyringController = await initializeKeyringController({
+            password: PASSWORD,
+            constructorOptions: {
+              cacheEncryptionKey: true,
+              encryptor: mockEncryptor,
+            },
+          });
+          const initialVault = keyringController.store.getState().vault;
+          const updatedVaultMock =
+            '{"vault": "updated_vault_detail", "salt": "salt"}';
+          const mockEncryptionResult = {
+            data: '0x1234',
+            iv: 'an iv',
+          };
+          sinon.stub(mockEncryptor, 'updateVault').resolves(updatedVaultMock);
+          sinon
+            .stub(mockEncryptor, 'encryptWithKey')
+            .resolves(mockEncryptionResult);
+
+          await keyringController.unlockKeyrings(PASSWORD);
+          const updatedVault = keyringController.store.getState().vault;
+
+          expect(initialVault).not.toBe(updatedVault);
+        });
+      });
+
       describe(`with cacheEncryptionKey = false`, () => {
         it('should update the vault', async () => {
           const mockEncryptor = new MockEncryptor();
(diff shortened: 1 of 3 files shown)


src/KeyringController.ts
```diff
@@ -930,6 +930,7 @@ class KeyringController extends EventEmitter {
 
     if (
       this.password &&
+      (!this.#cacheEncryptionKey || !encryptionKey) &&
```
mikesposito, line 933: This ensures that we don't do the whole `persistAllKeyrings` operation for nothing, as the vault will not get updated if we reuse the cached encryption key

src/KeyringController.test.ts
```diff
@@ -199,83 +211,86 @@ describe('KeyringController', () => {
...
+      describe('when `encryptionKey` is not set and `password` is set', () => {
+        it('should save an up to date encryption salt to the `memStore` when `password` is set through `createNewVaultAndKeychain`', async () => {
+          const keyringController = await initializeKeyringController({
+            password: PASSWORD,
+            constructorOptions: {
+              cacheEncryptionKey: true,
+            },
+          });
+          await keyringController.createNewVaultAndKeychain(PASSWORD);
+          deleteEncryptionKeyAndSalt(keyringController);
```
Gudahtt, line 261: Ah I see; this method is needed because this _kinda_ isn't a real use-case. Or it us, but only as part of the initial unlock/vault creation, it's never a use case when calling `persistAllKeyrings` directly.

Makes sense. Seems like a bit of a code smell that this condition is possible only when this external method is called in a certain way internally, but that problem will go away soon hopefully when we merge this into core.

src/KeyringController.test.ts
```diff
@@ -836,29 +851,63 @@ describe('KeyringController', () => {
...
-              '{"vault": "updated_vault_detail", "salt": "salt"}';
-            sinon.stub(mockEncryptor, 'updateVault').resolves(updatedVaultMock);
-            sinon.stub(mockEncryptor, 'encrypt').resolves(updatedVaultMock);
-
-            await keyringController.unlockKeyrings(PASSWORD);
-            const updatedVault = keyringController.store.getState().vault;
-
-            expect(initialVault).not.toBe(updatedVault);
-            expect(updatedVault).toBe(updatedVaultMock);
+      describe(`with cacheEncryptionKey = true and encryptionKey is unset`, () => {
```
Gudahtt, line 854: Nit: Is there a test case for when cacheEncryptionKey = true and encryptionKey is set?

Gudahtt (dismissed):
LGTM!

src/KeyringController.test.ts
Gudahtt: A nicer way to test this would be to test it as part of `createNewVaultAndKeychain` (and the other methods that trigger this condition), but given that this is to be deleted soon I don't think it'd be worth the effort :sweat_smile:

src/KeyringController.test.ts
mikesposito: Added in https://github.com/MetaMask/KeyringController/pull/307/commits/fe45b61ff02c4583aa2aaae3c931bc19b269ea64!

Gudahtt (approved):

PR #309, opened 2023-11-28
Increase unit test coverage

mikesposito:
## Description

This PR increases the unit test coverage of the controller. 

Current change:
```
branches    80.18 → 90.09  +9.91
functions   93.54 → 96.77  +3.23
lines       91.69 → 97.44  +5.75
statements  91.87 → 97.5  +5.63
```

## Changes

n/a

## References

n/a

## Checklist

### DIFF
--- a/src/KeyringController.test.ts
+++ b/src/KeyringController.test.ts
@@ -1,7 +1,7 @@
 import { Wallet } from '@ethereumjs/wallet';
 import HdKeyring from '@metamask/eth-hd-keyring';
 import { normalize as normalizeAddress } from '@metamask/eth-sig-util';
-import type { Hex } from '@metamask/utils';
+import type { Hex, Json, KeyringClass } from '@metamask/utils';
 import { bytesToHex } from '@metamask/utils';
 import { strict as assert } from 'assert';
 import * as sinon from 'sinon';
@@ -14,6 +14,11 @@ import {
   PASSWORD,
   MOCK_HARDCODED_KEY,
   MOCK_ENCRYPTION_SALT,
+  BaseKeyringMock,
+  KeyringMockWithDestroy,
+  baseTransactionMockFactory,
+  KeyringMockWithSignTransaction,
+  KeyringMockWithUserOp,
 } from './test';
 import type { KeyringControllerArgs } from './types';
 
@@ -60,7 +65,7 @@ async function initializeKeyringController({
   const keyringController = new KeyringController({
     encryptor: new MockEncryptor(),
     cacheEncryptionKey: false,
-    keyringBuilders: [keyringBuilderFactory(KeyringMockWithInit)],
+    keyringBuilders: [keyringBuilderFactory(BaseKeyringMock)],
     ...constructorOptions,
   });
 
@@ -93,6 +98,25 @@ function deleteEncryptionKeyAndSalt(keyringController: KeyringController) {
   delete keyringControllerState.encryptionKey;
   delete keyringControllerState.encryptionSalt;
   keyringController.memStore.updateState(keyringControllerState);
+}
+
+/**
+ * Stub the `getAccounts` and `addAccounts` methods of the given keyring class to return the given
+ * account.
+ *
+ * @param keyringClass - The keyring class to stub.
+ * @param account - The account to return.
+ */
+function stubKeyringClassWithAccount(
+  keyringClass: KeyringClass<Json>,
+  account: string,
+) {
+  sinon
+    .stub(keyringClass.prototype, 'getAccounts')
+    .returns(Promise.resolve([account]));
+  sinon
+    .stub(keyringClass.prototype, 'addAccounts')
+    .returns(Promise.resolve([account]));
 }
 
 describe('KeyringController', () => {
@@ -152,9 +176,12 @@ describe('KeyringController', () => {
     it('calls keyring optional destroy function', async () => {
       const keyringController = await initializeKeyringController({
         password: PASSWORD,
+        constructorOptions: {
+          keyringBuilders: [keyringBuilderFactory(KeyringMockWithDestroy)],
+        },
       });
-      const destroy = sinon.spy(KeyringMockWithInit.prototype, 'destroy');
-      await keyringController.addNewKeyring('Keyring Mock With Init');
+      const destroy = sinon.spy(KeyringMockWithDestroy.prototype, 'destroy');
+      await keyringController.addNewKeyring('Keyring Mock With Destroy');
 
       await keyringController.setLocked();
 
@@ -631,6 +658,9 @@ describe('KeyringController', () => {
     it('should call init method if available', async () => {
       const keyringController = await initializeKeyringController({
         password: PASSWORD,
+        constructorOptions: {
+          keyringBuilders: [keyringBuilderFactory(KeyringMockWithInit)],
+        },
       });
       const initSpy = sinon.spy(KeyringMockWithInit.prototype, 'init');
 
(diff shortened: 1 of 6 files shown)


Gudahtt (approved):
LGTM!

Gudahtt (dismissed):

src/test/transaction.mock.ts
```diff
@@ -0,0 +1,10 @@
+import { TransactionFactory, type TxData } from '@ethereumjs/tx';
+
+export const baseTransactionMockFactory = (options: TxData = {}) =>
```
mcmire, line 3: Nit: While this name is accurate — it is indeed a factory — it seems a bit unusual the name of this variable is a noun and not a verb phrase. What do you think about `buildMockTransaction`? Then again, if this is intended to mirror `keyringBuilderFactory`, then that's fine.

src/KeyringController.test.ts
```diff
@@ -1075,6 +1121,19 @@ describe('KeyringController', () => {
         'someapp.origin.io',
       ]);
     });
+
+    it('should throw error if keyring for account does not support getAppKeyAddress', async () => {
```
mcmire, line 1125: Nit: It looks like we aren't using "should" in the names of these tests.

So, for example:

``` suggestion
    it('throws error if keyring for account does not support getAppKeyAddress', async () => {
```

but a similar suggestion would apply to the other tests in this file.

src/KeyringController.test.ts
```diff
@@ -1356,10 +1430,91 @@ describe('KeyringController', () => {
...
+      expect(signTransactionSpy.calledOnce).toBe(true);
+      expect(signTransactionSpy.getCall(0).args[0]).toStrictEqual(address);
+      expect(signTransactionSpy.getCall(0).args[1]).toStrictEqual(
+        mockTransaction,
+      );
+    });
+  });
+
+  describe('signMessage', () => {
+    it('should sign message', async () => {
```
mcmire, line 1517: Can we go into a little more detail on what the expected behavior here is? What does "signing" mean? It looks like the input data is encrypted using the seed phrase? Does `signMessage` on the keyring do most of the work? If so, it might be important to highlight the keyring being used in this test, since the exact keyring being used affects the signing.

src/KeyringController.test.ts
```diff
@@ -834,6 +867,19 @@ describe('KeyringController', () => {
       // removing the account since it still has an account left
       expect(keyringController.keyrings).toHaveLength(2);
     });
+
+    it('throws an error if the keyring for the given account does not support account removal', async () => {
+      const mockAccount = '0x5aC6d462f054690A373Fabf8cc28E161003aEB19';
+      stubKeyringClassWithAccount(BaseKeyringMock, mockAccount);
+      const keyringController = await initializeKeyringController({
+        password: PASSWORD,
+      });
```
mcmire, line 876: I don't know if this is a pre-existing pattern in these tests, but there seems to be a disconnect in the described behavior, the data that the test is using, and the place where that data is defined. This test seems to be relying on the fact that BaseKeyringMock is the default keyring for all tests. Defaults are generally useful for data that doesn't matter, but in this case the keyring _does_ matter, as the test will only pass when using a keyring that doesn't support account removal.

Maybe just to be absolutely certain that the test does the right thing, it would be useful to pass BaseKeyringMock as the keyring used?

``` suggestion
      stubKeyringClassWithAccount(BaseKeyringMock, mockAccount);
      const keyringController = await initializeKeyringController({
        password: PASSWORD,
        constructorOptions: {
          keyringBuilders: [keyringBuilderFactory(BaseKeyringMock)],
        }
      });
```

That way it's easier to see what the controller is using and why the error below is produced.

Again, though, if this would require refactoring these tests to do, don't worry about it — just something that might improve the readability of these tests in the future.

src/KeyringController.test.ts
```diff
@@ -1376,21 +1531,36 @@ describe('KeyringController', () => {
...
+
+      await expect(keyringController.signMessage(inputParams)).rejects.toThrow(
+        KeyringControllerError.UnsupportedSignMessage,
+      );
+    });
+  });
 
-      await keyringController.addNewKeyring('Keyring Mock With Init');
+  describe('prepareUserOperation', () => {
+    it('prepares a base UserOperation', async () => {
```
mcmire, line 1554: Since the name of this test simply repeats the name of the method, are there any more details we can supply here? I'm noticing that the JSDocs for this method say "Convert a base transaction to a base UserOperation", so although it's a bit jargony, it paints a slightly better picture:

``` suggestion
    it('converts a base transaction to a base UserOperation', async () => {
```

or if we want to be a bit more literal:

``` suggestion
    it('calls the prepareUserOperation method on the keyring, returning its result', async () => {
```

src/KeyringController.test.ts
```diff
@@ -1421,13 +1593,41 @@ describe('KeyringController', () => {
...
+      );
+
+      await expect(result).rejects.toThrow(
+        'KeyringController - The keyring for the current address does not support the method prepareUserOperation.',
+      );
+    });
+  });
+
+  describe('patchUserOperation', () => {
+    it('patches an UserOperation', async () => {
```
mcmire, line 1621: Similar as above — perhaps:

``` suggestion
    it('uses the patchUserOperation method on the keyring to patch properties of a UserOperation', async () => {
```

or, more literally:

``` suggestion
    it('calls the patchUserOperation method on the keyring, returning its result', async () => {

File: src/KeyringController.test.ts [hunk 7]
```diff
@@ -1479,50 +1673,28 @@ describe('KeyringController', () => {
...
       await expect(result).rejects.toThrow(
-        'KeyringController - The keyring for the current address does not support the method prepareUserOperation.',
+        'KeyringController - The keyring for the current address does not support the method patchUserOperation.',
       );
     });
+  });
 
-    it("throws when the keyring doesn't implement patchUserOperation", async () => {
+  describe('signUserOperation', () => {
+    it('signs an UserOperation', async () => {
```
mcmire, line 1688: Similar as above:

``` suggestion
    it('calls the signUserOperation method on the keyring, returning its result', async () => {

mcmire (commented):
Looks good, thanks for doing this. All of my comments are minor, but it might be nice to be a bit more descriptive with the test names.

src/test/transaction.mock.ts
mikesposito: Indeed I initially named it after the `keyringBuilderFactory`, but what you are saying makes more sense - `buildMockTransaction` sounds good to me, refactored in 988974e

src/KeyringController.test.ts
mikesposito: Good point, the pattern is not clear in the file: some tests use "should", while others use "throws, forwards, adds, etc.."

I changed all the occurrences that I found, removing "should" in 988974e

src/KeyringController.test.ts
mikesposito: Good point, the test relies on `BaseKeyringMock` as the default "additional" keyring during tests (in the `initializeKeyringController` function), but this might be a hidden behavior that makes the test case less readable.

I applied your suggestion in 988974e

src/KeyringController.test.ts
mikesposito: The `data` is usually a digest of the message encoded as hex (so it is not encrypted, just hashed with keccak256), ready to be signed with an ecdsa private key, that in our case is handled by the Keyring linked to the `from` account.

So, the target Keyring must indeed support the `signMessage` method, and the "real work" is done by it.

Because of this, I would be inclined to change it to `forwards the message to the keyring for signing` (using  `signTransaction` tests as an example) - but on the other hand this test case doesn't only check if the message gets forwarded to the Keyring: it also checks for the signature correctness, and this is the reason why we use a seed phrase with a known and predictable first-account-address

mcmire: Gotcha, yeah that's a bit tricky. Okay, perhaps we can think of a better name later.

mcmire (approved):

PR #310, opened 2023-11-28
Update `browser-passworder` to `^4.3.0`

mikesposito:
## Description

This PR updates `@metamask/browser-passworder` minimum version to 4.3.0

The library `4.3.0` version also provides `isVaultUpdated`, which gives us a chance to make the way `EthKeyringController` checks for a potential encryption upgrade more efficient.

## Changes

- **ADDED**: Optional `isVaultUpdated` property to `GenericEncryptor` type
- **CHANGED**: Bump `@metamask/browser-passworder` to `^4.3.0`
- **CHANGED**: When the encryptor provides an `isVaultUpdated` method, the latter is used to check if the vault is updated instead of `updateVault`
- **REMOVED**: `GenericEncryptor`'s `updateVault` property has been removed as no longer used 

## References

* See: https://github.com/MetaMask/core/issues/3582

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "@ethereumjs/tx": "^4.2.0",
-    "@metamask/browser-passworder": "^4.2.0",
+    "@metamask/browser-passworder": "^4.3.0",
     "@metamask/eth-hd-keyring": "^7.0.1",
     "@metamask/eth-sig-util": "^7.0.0",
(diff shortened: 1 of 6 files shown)


Gudahtt (approved):
LGTM! Though we should mention the performance improvement in the changelog as well, not just the API change

Gudahtt (approved):

PR #315, opened 2023-12-01
feat!: add methods to support ERC-4337 accounts

danroc:
## Description

**Depends on:**

- [x] <https://github.com/MetaMask/keyring-api/pull/213>

This PR adds the following new methods available in the `EthKeyring` interface:

- `prepareUserOperation`
- `patchUserOperation`
- `signUserOperation`

See the [PR](https://github.com/MetaMask/keyring-api/pull/213) in the `keyring-api` for more details.

**BREAKING:** This PR replaces the `Keyring` interface from `@metamask/utils` with the `EthKeyring` interface from `@metamask/keyring-api`.

## Checklist


danroc:
@SocketSecurity ignore @ethereumjs/rlp@4.0.1
@SocketSecurity ignore @ethereumjs/rlp@5.0.1
@SocketSecurity ignore rlp@2.2.7

These are transient dependencies and this script isn't used.

src/KeyringController.ts
```diff
@@ -30,20 +36,21 @@ import type {
...
+import { throwError } from './utils';
 
 const defaultKeyringBuilders = [
   keyringBuilderFactory(SimpleKeyring),
   keyringBuilderFactory(HDKeyring),
 ];
 
 class KeyringController extends EventEmitter {
-  keyringBuilders: { (): Keyring<Json>; type: string }[];
+  keyringBuilders: { (): EthKeyring<Json>; type: string }[];
```
gantunesr, line 47: Since there's a type change from `Keyring` to `EthKeyring` it'd be good to indicate the breaking change in the PR description

src/KeyringController.ts
```diff
@@ -579,6 +586,64 @@ class KeyringController extends EventEmitter {
...
+   *
+   * @param rawAddress - Address of the sender.
+   * @param transactions - Base transactions to include in the UserOperation.
+   * @returns A pseudo-UserOperation that can be used to construct a real.
+   */
+  async prepareUserOperation(
+    rawAddress: string,
+    transactions: EthBaseTransaction[],
+  ): Promise<EthBaseUserOperation> {
+    const address = normalizeToHex(rawAddress) as Hex;
```
gantunesr, line 600: Isn't `normalizeToHex ` already returning a `Hex`? Not sure why the type casting is required here and the below methods

gantunesr (commented):
LGTM! Left some minor comments to be resolved before approving

package.json
```diff
@@ -72,7 +73,7 @@
     "depcheck": "^1.4.7",
     "eslint": "^8.48.0",
     "eslint-config-prettier": "^8.7.0",
-    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-import": "~2.26.0",
```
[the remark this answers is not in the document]
danroc: Yarn was complaining that the version didn't satify the [requirements](https://github.com/MetaMask/eslint-config/blob/3248a64a63cd7fbc500b9836975489c7f1be8e0b/package.json#L33) from `@metamask/eslint-config`.

src/KeyringController.ts
danroc: Actually `normalizeToHex` has the following signature:

```ts
function normalize(input: number | string): string | undefined;
```

I kept the same pattern used in the other signing methods above, but I agree that it's not ideal. We should probably update the `normalize` method from `@metamask/eth-sig-utils` to always return a `Hex`.

src/KeyringController.ts
danroc: Indeed, good point.

gantunesr (approved):

PR #329, opened 2024-01-04
feat!: create new vault with any keyring

gantunesr:
## Description

This change updates the logic inside the `eth-keyring-controller` to allow the creation of a vault with any keyring. This removes the hardcoded HD Keyring from the code and allow the client to pass the parameters to instantiate any keyring that is accepted by the controller.

## Changes

- **BREAKING**: Unify `createNewVaultAndKeychain` and `createNewVaultAndRestore` into new method `createNewVaultWithKeyring`. `createNewVaultWithKeyring` accepts a password and a keyring object provided by the client and returns the `KeyringControllerState`

## References

No references

## Checklist

### DIFF
--- a/src/KeyringController.ts
+++ b/src/KeyringController.ts
@@ -115,43 +115,25 @@ class KeyringController extends EventEmitter {
    */
 
   /**
-   * Create New Vault And Keychain
+   * Create new vault And with a specific keyring
    *
    * Destroys any old encrypted storage,
    * creates a new encrypted store with the given password,
-   * randomly creates a new HD wallet with 1 account,
-   * faucets that account on the testnet.
+   * creates a new wallet with 1 account.
    *
    * @fires KeyringController#unlock
    * @param password - The password to encrypt the vault with.
+   * @param keyring - A object containing the params to instantiate a new keyring.
+   * @param keyring.type - The keyring type.
+   * @param keyring.opts - Optional parameters required to instantiate the keyring.
    * @returns A promise that resolves to the state.
    */
-  async createNewVaultAndKeychain(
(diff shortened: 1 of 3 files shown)


src/KeyringController.ts
```diff
@@ -115,60 +115,33 @@ class KeyringController extends EventEmitter {
...
-   * @param password - The password to encrypt the vault with.
-   * @param seedPhrase - The BIP39-compliant seed phrase,
-   * either as a string or Uint8Array.
-   * @returns A promise that resolves to the state.
-   */
-  async createNewVaultAndRestore(
-    password: string,
-    seedPhrase: Uint8Array | string | number[],
+    keyring: {
+      type: string;
```
owencraston, line 134: is there now better way we can type the options here?

owencraston: should we default to HD if one is not passed in?

gantunesr: > _is there now better way we can type the options here? I see in the tests we are using KeyringType?_


No, some keyrings are passed from the outside (`QRKeyring`, `LedgerLeyring`, `SnapKeyring`, etc) and there isn't a way to know what they would be unless we enforce it, for the moment having them as `unknown` and having the keyring itself handle it is the best choice.

> _should we default to HD if one is not passed in?_

I don't see the benefit of it at least inside this package, the clients should handle it

mikesposito (approved):
Looks good!

Question: will this be done in `@metamask/keyring-controller` too?

gantunesr:
@mikesposito Yes! I'll have this done in @metamask/keyring-controller too

PR #366, opened 2024-03-05
BREAKING: Increase minimum node version to 18

legobeat:
## Description

- Remove support for major Node.js versions 16,17,19
- Update recommended Node version in readme
- ci: bump default version from 16 to 18
- ci: remove node-version 16 from matrix

## Changes

- **BREAKING**: Drop support for Node.js versions 16, 17, 19

## Checklist


mikesposito (approved):

mcmire:
I don't think we were planning another major release of this package as it's in maintenance mode and has been replaced by `@metamask/keyring-controller`. Do we think this is necessary from a security perspective though?

